PR #4, opened 2014-08-28
In product.price.type do not set company_id required

laetitia-gangloff:
I set company_id optional.

I think, the context can be sufficient to know which price take. With a required company_id it is difficult to use pricelist inheritence.

For example a base price list for all my company which contains product with same price for all. And one pricelist per company with specific price for product or based on the general pricelist.

### DIFF
--- a/product_price_history/product_price_history.py
+++ b/product_price_history/product_price_history.py
@@ -323,8 +323,7 @@ class price_type(orm.Model):
     _inherit = "product.price.type"
 
     _columns = {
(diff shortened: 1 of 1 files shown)


lmignon:
:+1: LGTM; failing tests seems to be unrelated

rvalyi:
:+1:  LGTM

pedrobaeza:
Security rule it's also valid for this condition, so :+1:

laetitia-gangloff:
Thank you !

Issue #6, opened 2014-10-30
`product_get_cost_field` and `stock` both use "Cost price" to name different fields


clonedagain:
As a workaround for https://github.com/odoo/odoo/issues/3396, I think the module `product_get_cost_field` should change the label of `product.product.standard_price` to something else but "Cost price", to make it clear what each field does.

PS: as it turns out, the wording in standard module `stock` is probably right and `product_get_cost_field` should not add a field labeled "Standard Price".
To avoid breaking existing installations, in the short term the field `cost_price` should be relabeled to another name.
In the longer term, we should strive to integrate `product_get_cost_field`'s features deeper into the standard costing features.

clonedagain:
You can assign this to me unless someone objects to it.

yvaucher:
This module was developed to know the cost of a product to make margin analysis. So cost price seems suited. Maybe as you said we need 2 different fields. For inventory valuation, the price might be different than purchase price.

I would prefer a name like 'inventory_value' as 'value' seems too generic.

clonedagain:
Yes sorry if this is not clear, I'm proposing to rename the standard's 
`standard_price` to "Value" (or "Inventory Value", it sounds better indeed).
`cost_price` must indeed be named "Cost Price".

clonedagain:
From https://github.com/odoo/odoo/issues/3396#issuecomment-61172164 I gather that the change must be made the other way round: Cost Price is the right term for inventory valuation, and we have to name our field something else.

clonedagain:
@yvaucher can you think of another name instead of "cost price" for OCA's module?

clonedagain:
After reading a bit more on the subject, I'm not quite sure we really should have 2 fields.
For example in the classic book "Gestion industrielle", Blondel writes:
"La logique économique et les règles fiscales impliquent la comptabilisation de coûts lorsqu’ils ont été constatés. ... Le mode de valorisation des composants n’est pas neutre sur la structure de coût de revient. Il sera cohérent avec la méthode adoptée pour la valorisation des stocks"
(rough translation: Economic logic and the tax rules involve cost accounting when they are recorded. ... The valuation mode of components is not neutral on the structure of cost. It will be consistent with the method adopted for the valuation of stocks)

So having 2 distinct fields may actually be bad practice, if not downright forbidden.

[...]

PR #8, opened 2014-10-30
Change labels to distinguish Cost Price from Inventory Valuation

clonedagain:
As a workaround for odoo/odoo#3396, I think the module product_get_cost_field should change the label of product.product.standard_price to something else but "Cost price", to make it clear what each field does.
I propose to name it "Inventory Value" since it's used for stock valuation.

I also simplified the view and added a bit of documentation.

capture du 2014-10-30 17 38 23

### DIFF
--- a/product_cost_incl_bom/product_cost_incl_bom.py
+++ b/product_cost_incl_bom/product_cost_incl_bom.py
@@ -329,10 +329,8 @@ class product_product(orm.Model):
         'cost_price': fields.function(
             _cost_price,
             store=_cost_price_triggers,
-            string='Cost Price (incl. BoM)',
+            string='Cost Price',
             digits_compute=dp.get_precision('Product Price'),
-            help="The cost price is the standard price or, if the "
-                 "product has a bom, the sum of all standard price "
-                 "of its components. it take also care of the bom "
-                 "costing like cost per cycle.")
+            help="The cost that you have to support in order to produce or "
+                 "acquire the goods.")
     }
--- a/product_get_cost_field/product_get_cost_field.py
+++ b/product_get_cost_field/product_get_cost_field.py
@@ -25,10 +25,12 @@ import decimal_precision as dp
 _logger = logging.getLogger(__name__)
 
 
-class product_product(orm.Model):
+class ProductProduct(orm.Model):
     _inherit = 'product.product'
 
     def _compute_purchase_price(self, cr, uid, ids, context=None):
+        """"Use standard_price's value for cost_price"""
+        # TODO: remove in v8
         res = {}
         if isinstance(ids, (int, long)):
             ids = [ids]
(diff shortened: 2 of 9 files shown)


product_get_cost_field/product_get_cost_field.py
```diff
@@ -39,21 +41,23 @@ def _compute_purchase_price(self, cr, uid, ids, context=None):
         return res
 
     def _cost_price(self, cr, uid, ids, field_name, arg, context=None):
-        if context is None:
-            context = {}
+        '''Proxy method to make the function field modular
```
pedrobaeza: Use double quotes for docstring.

product_get_cost_field/product_get_cost_field.py
```diff
@@ -25,10 +25,12 @@
 _logger = logging.getLogger(__name__)
 
 
-class product_product(orm.Model):
+class ProductProduct(orm.Model):
     _inherit = 'product.product'
 
     def _compute_purchase_price(self, cr, uid, ids, context=None):
+        '''Use standard_price's value for cost_price'''
```
pedrobaeza: Use double quotes for docstring.

product_get_cost_field/product_get_cost_field.py
```diff
@@ -39,21 +41,23 @@ def _compute_purchase_price(self, cr, uid, ids, context=None):
...
 
     def get_cost_field(self, cr, uid, ids, context=None):
         return self._cost_price(cr, uid, ids, '', [], context=context)
 
     def _get_product_from_template(self, cr, uid, ids, context=None):
-        prod_obj = self.pool.get('product.product')
-        prod_ids = prod_obj.search(cr, uid,
-                                   [('product_tmpl_id', 'in', ids)],
-                                   context=context)
+        '''Find the products to trigger when a template changes'''
```
pedrobaeza: Double quotes.

product_get_cost_field/product_get_cost_field.py
```diff
@@ -69,6 +73,6 @@ def _get_product_from_template(self, cr, uid, ids, context=None):
             store=_cost_price_triggers,
             string='Cost Price',
             digits_compute=dp.get_precision('Product Price'),
-            help="The cost price is the standard price unless you install the "
-                 "product_cost_incl_bom module.")
+            help="the cost that you have to support in order to produce or "
```
pedrobaeza: s/the/The

pedrobaeza:
Minor issues. The rest LGTM.

clonedagain:
@pedrobaeza so you agree to the principle?

pedrobaeza:
Yeah, of course. Terms must be adjusted, and until we get in on upstream, we can change them here.

product_get_cost_field/product_get_cost_field.py
```diff
@@ -39,21 +41,24 @@ def _compute_purchase_price(self, cr, uid, ids, context=None):
...
 
     def _get_product_from_template(self, cr, uid, ids, context=None):
-        prod_obj = self.pool.get('product.product')
-        prod_ids = prod_obj.search(cr, uid,
-                                   [('product_tmpl_id', 'in', ids)],
-                                   context=context)
+        """Find the products to trigger when a template changes"""
+        # self could be product template so we'll lookup the pool
+        prod_ids = self.pool['product.product'].search(
+            cr, uid, [('product_tmpl_id', 'in', ids)], context=context)
```
clonedagain, line 58: previous version of this branch proposed self.search but it doesn't work (somehow self was "product.template" during write() ), so I've commented about it instead

product_cost_incl_bom/product_cost_incl_bom.py
```diff
@@ -329,10 +329,8 @@ def _get_product_from_template2(self, cr, uid, ids, context=None):
...
             _cost_price,
             store=_cost_price_triggers,
-            string='Cost Price (incl. BoM)',
+            string='Cost Price',
             digits_compute=dp.get_precision('Product Price'),
-            help="The cost price is the standard price or, if the "
-                 "product has a bom, the sum of all standard price "
-                 "of its components. it take also care of the bom "
-                 "costing like cost per cycle.")
+            help="The cost that you have to support in order to produce or "
```
clonedagain, line 334: Here I keep the strings the same as in  `product_get_cost_field` because exporting the POT misses the translations for those strings.

product_get_cost_field/product_get_cost_field.py
pedrobaeza, line 58: Yeah, it can be a bad super call in the chain due to the duality product.template/product.product

clonedagain:
Abandoning for now, as per https://github.com/OCA/margin-analysis/issues/6#issuecomment-61190689.
I'll try to come up with something else in the coming days.

PR #10, opened 2014-11-05
[IMP] Rename cost_price to 'Replenishment cost'

### LINKED ISSUE #6: `product_get_cost_field` and `stock` both use "Cost price" to name different fields

clonedagain:
Fixes #6
This change is meant as a short-term solution, to avoid confusion between standard_price and cost_price since both are named 'Cost Price' in the GUI.
The accounting best practises seem to be that if a Cost price is known, it must be used for the valuation of the stock. So the right 'Cost price' field is standard_price and we have to name cost_price something else.
The description of the inherited fields is changed too, because for some reason the server seems unable to export the new names in the .po files.

### DIFF
--- a/product_cost_incl_bom/__openerp__.py
+++ b/product_cost_incl_bom/__openerp__.py
@@ -18,8 +18,8 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-{'name': 'Product Cost incl. BOM',
- 'version': '1.0',
+{'name': 'Replenishment Cost incl. BOM',
+ 'version': '1.1',
  'author': 'Camptocamp',
  'maintainer': 'Camptocamp',
  'category': 'Products',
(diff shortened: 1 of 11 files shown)


pedrobaeza:
I'm still with you :+1:

product_cost_incl_bom/__openerp__.py
```diff
@@ -31,11 +31,11 @@
 Product Cost including BOM costs
```
yvaucher: I think this title could go in `summary` and it could be replaced by Manufacturing cost

clonedagain:
After more thinking (really I can never get enough thinking sorry), I think it should actually be called "Replenishment Cost", because that would let me cleanly introduce a module to optionally base the value on purchases instead of standard prices.
@pedrobaeza would that suit you?

clonedagain:
@yvaucher I factored your comment in too.

pedrobaeza:
OK, seems good. Please put a big help about the use of the field.

clonedagain:
I've extended the help text, does it suit you now please?

clonedagain:
I've pushed an early version of the extension I'd like to propose which requires the name change in #12.

pedrobaeza:
It looks good to me, :+1:, waiting only that Travis doesn't say the contrary.

clonedagain:
not sure whether the remaining travis errors are related or not...

pedrobaeza:
Yeah, Travis problems are unrelated.

guewen:
:+1: 

Thanks!

PR #12, opened 2014-11-14
[WIP] replenishment cost based on purchases

clonedagain:
The first commit hacks it's way around the limitations related to function fields and _inherit.
This is still work in progress, it seems to work but automatic tests are not finished yes.
Early opinons about this feature are welcome however.

### DIFF
--- a/product_cost_incl_bom/__openerp__.py
+++ b/product_cost_incl_bom/__openerp__.py
@@ -26,6 +26,8 @@
  'complexity': "normal",  # easy, normal, expert
  'depends': ['product_get_cost_field',
              'mrp',
+             #XXX OK so we're just hijacking the whole thing...
+             'product_cost_purchase',
              ],
  'description': """
 Replenishment Cost including BOM costs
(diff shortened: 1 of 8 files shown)


product_cost_incl_bom/__openerp__.py
```diff
@@ -18,24 +18,26 @@
...
+{'name': 'Replenishment Cost incl. BOM',
+ 'version': '1.1',
  'author': 'Camptocamp',
  'maintainer': 'Camptocamp',
  'category': 'Products',
  'complexity': "normal",  # easy, normal, expert
  'depends': ['product_get_cost_field',
              'mrp',
+             #XXX OK so we're just hijacking the whole thing...
+             'product_cost_purchase',
```
clonedagain, line 30: I'd really like to find _another way_ to make sure the methods of `product_cost_purchase` are called _before_ those of `product_cost_incl_bom`.
This a blocker for this proposal, we can't force existing users to use a new feature they may not want.

clonedagain:
Abandoned project for some time already, closing.

PR #24, opened 2015-07-02
[7.0] 'product_standard_margin' margin stored

legalsylvain:
Make the field 'standard_margin' stored, depending of product fields and taxes;

So, this PR allows to filter or order more easely by margin, that is a interesting need.


pedrobaeza:
Looks good :+1:

rvalyi:
:+1:  LGTM

PR #28, opened 2015-09-11
BoM price history: correct call to store function

mdietrichc2c:
PR to fix an issue with the store condition: the super called was `mrp.bom.get_product` , but the function in `product_cost_incl_bom` is on the `product.product` object.


guewen:
:+1:

yvaucher:
:+1:

pedrobaeza:
:+1:

PR #30, opened 2016-04-22
[MIG][8.0] replenishment_cost + product_cost_incl_bom

SodexisTeam:


atchuthan:
@pedrobaeza: add "needs review" label for this PR.

gurneyalex:
:+1:

README.md
```diff
@@ -15,14 +15,14 @@ Available addons
...
 --- | --- | ---
+[product_cost_incl_bom](product_cost_incl_bom/) | 8.0.2.1.0 | Product Cost incl. BOM
 [product_replenishment_cost](product_replenishment_cost/) | 8.0.2.0.0 | Replenishment cost
 [product_standard_margin](product_standard_margin/) | 8.0.2.0.0 | Product Margin and Margin Rate
 
 Unported addons
 ---------------
 addon | version | summary
 --- | --- | ---
-[product_cost_incl_bom](product_cost_incl_bom/) | 1.0 (unported) | Product Cost incl. BOM
```
Garamotte: This file is updated automatically by an OCA script.

product_cost_incl_bom/models/product.py
```diff
@@ -0,0 +1,217 @@
...
+        return computed
+
+    @api.multi
+    def _get_replenishment_cost(self):
+        res = self._compute_purchase_price()
+        for product in self:
+            cost = 0
+            if product.id in res.keys():
+                cost = res[product.id]
+            product.replenishment_cost = cost
```
Garamotte: Maybe `product.replenishment_cost = res.get(product.id, 0)` ?

product_replenishment_cost/models/__init__.py
```diff
@@ -1,22 +1,6 @@
...
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# © 2012 Alexandre Fayolle,Yannick Vaucher,Joël Grand-Guillaume,Camptocamp
+# © 2016 Sodexis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from . import product_product
+from . import product_product, product_template
```
Garamotte: Please split one import per line.

product_replenishment_cost/models/product_product.py
```diff
@@ -28,14 +12,15 @@
...
-    @api.one
+    @api.multi
     @api.depends('product_tmpl_id.standard_price')
     def _get_replenishment_cost(self):
-        self.replenishment_cost = self.standard_price
+        for product in self:
+            product.replenishment_cost = product.standard_price
 
     replenishment_cost = fields.Float(
-        compute=_get_replenishment_cost, store=True,
```
Garamotte, line 37: Why did you remove the `store` attribute ?

product_replenishment_cost/models/product_template.py
```diff
@@ -0,0 +1,32 @@
...
+import openerp.addons.decimal_precision as dp
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    @api.multi
+    @api.depends('standard_price')
+    def _get_replenishment_cost(self):
+        replenishment_cost = 0.0
```
Garamotte: If a template has no variant, it will pick the value computed for the previous template in the loop.
You should initialize the variable in the for loop.

product_replenishment_cost/models/product_template.py
```diff
@@ -0,0 +1,32 @@
...
+        for template in self:
+            variants = template.mapped('product_variant_ids')
```
Garamotte: `variants = template.product_variant_ids` is enough here, mapped is useful when you work on multiple records, or through multiple fields.

Garamotte (changes requested):

product_replenishment_cost/models/product_product.py
atchuthan: I had 2 reasons to remove store=True
1. The replenishment cost field defined in product_replenishment_cost module with store=True, creates the field in database with the value of product's standard_price and the value is not recomputed at the time of product_cost_incl_bom module installation.
2. Replenishment cost needs to be changed for each change in bom lines of product, when there is a change in their replenishment cost value and it did not work in that case as the method was not triggered for store=True. (when we change the BOM directly, it did not trigger the compute method and it does not allow multi-level BOM)

atchuthan:
@sylvain-garancher corrections done

gurneyalex (approved):

dvdhinesh (approved):
Code review :+1:

product_replenishment_cost/models/product_product.py
atchuthan: @sylvain-garancher please check the points I have listed out.

Garamotte: Ok, I see, and I don't find a trigger for that. With the explanation, it's fine.

If the issue is with the other module, maybe you can leave `store=True` here, and explicitely change to `store=False` in the other module ? (with a comment explaining why)

guewen:
Restarted Travis and I don't get why the builds are shown as failed whereas the builds exit with 0.

legalsylvain:
closing.
replenishment_cost is available in V8 and in V10.

You can also review the PR for V12 here : https://github.com/OCA/margin-analysis/pull/50

PR #33, opened 2017-05-18
Add sale_line_cost_control

guewen:
This module improves cost controlling on sales order lines.
Specifically, it:

* adds a checkbox on the lines indicating if the cost is missing on the product
* shows the linked product's standard price on the details of a line (form view)
* adds a button to set a cost price on the product directly from the sale line


sale_line_cost_control/models/sale.py
```diff
@@ -0,0 +1,31 @@
...
+                line.purchase_price,
+                precision_digits=precision
+            )
+
+    @api.multi
+    def action_set_purchase_price_on_line(self):
+        self.ensure_one()
+        action_xmlid = ('sale_line_cost_control.'
+                        'action_set_sale_line_purchase_price')
+        return self.env.ref(action_xmlid).read()[0]
```
gurneyalex: why use an object method rather than the action in the view?

sale_line_cost_control/wizards/set_sale_line_purchase_price.py
```diff
@@ -0,0 +1,53 @@
...
+        string='Cost',
+        digits=dp.get_precision('Product Price'),
+        groups="base.group_user",
+        help="Cost of the product template used for standard stock valuation "
+             "in accounting and used as a base price on purchase orders. "
+             "Expressed in the default unit of measure of the product.")
+    currency_id = fields.Many2one(
+        comodel_name='res.currency',
+        string="Currency",
+        required=True,
```
gurneyalex, line 29: do we want to default to the company currency ?

gurneyalex (approved):
globally approve

sale_line_cost_control/wizards/set_sale_line_purchase_price.py
guewen: Currently it's the sale order's currency. But now you tell it, I tend to think it makes more sense to use the company one...

sale_line_cost_control/models/sale.py
guewen: Previously I was modifying the context before returning it that's why I used a method. Now I removed the context keys, so indeed there is no particular reason to use a method.

guewen:
@gurneyalex modified my PR according to your suggestions

damdam-s (approved):
thanks

sale_line_cost_control/tests/common.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
```
simahawk, line 1: duplicated header

sale_line_cost_control/wizards/set_sale_line_purchase_price_views.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+  <record id="view_set_sale_line_purchase_price" model="ir.ui.view">
+    <field name="name">Set Cost</field>
+    <field name="model">set.sale.line.purchase.price</field>
+    <field name="arch" type="xml">
+      <form string="Set Cost">
+        <group>
```
simahawk: you might want to name the group to ease customization

simahawk (approved):
LGTM

PR #38, opened 2018-07-10
[ADD] sale_report_margin: New Module

chienandalu:
Adds `purchase_price` field to sale report.

cc @Tecnativa


legalsylvain (approved):

sale_report_purchase_price/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Adds `purchase_price` field to sale report.
```
pedrobaeza: Add here an explanation of what this field means (which is not the real purchase price).

pedrobaeza (changes requested):
I think we should go for the name `sale_report_margin`, and put on the roadmap the possibility of adding the gross margin and comparisons.

chienandalu:
@pedrobaeza Changes were already done

pedrobaeza (approved):

PR #39, opened 2018-07-13
[ADD] sale_margin_delivered: New Module

chienandalu:
Computes sale order lines margins for the delivered items.

cc  @Tecnativa


sale_margin_delivered/models/stock_picking.py
```diff
@@ -0,0 +1,20 @@
...
+    @api.multi
+    def action_done(self):
+        res = super().action_done()
+        # Update purchase_price in so_lines
+        pickings = self.filtered(
+            lambda x: x.picking_type_code == 'outgoing')
+        so_lines = pickings.mapped(
+            'move_lines.sale_line_id')
+        for line in so_lines:
+            line.purchase_price = line.product_id.standard_price
```
pedrobaeza: Why this code, @sergio-teruel ? This is modifying the behavior and I'm not sure this fits in this module.

sale_margin_delivered/views/sale_margin_delivered_view.xml
```diff
@@ -0,0 +1,19 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3). -->
+<odoo>
+
+<record id="view_order_form" model="ir.ui.view" >
+    <field name="model">sale.order</field>
+    <field name="inherit_id" ref="sale.view_order_form"/>
+    <field name="groups_id" eval="[(4,ref('sale_margin_security.group_sale_margin_security'))]"/>
+    <field name="arch" type="xml">
+       <xpath expr="//field[@name='order_line']/tree//field[@name='tax_id']" position="before">
+           <field name="margin" string="Margin"/>
```
pedrobaeza, line 12: Isn't this field already present in standard `sale_margin`?

chienandalu: Only globally, not per line

CasVissers-360ERP (approved):
Nice module!

sale_margin_delivered/models/stock_picking.py
pedrobaeza: Can you please answer this, @sergio-teruel ?

CasVissers-360ERP: @pedrobaeza I think this is because when using average or FiFo cost price method the standard price can change between moment of order & moment of delivery. When analyzing the margin on deliveries you want to take the cost price at delivery, not at order. The issue I might see is backorders etc.

pedrobaeza: Exactly, you're modifying a price that can be used for several deliveries. IMO, you should leave the one of the purchase confirmation (that's to say, remove this code).

CasVissers-360ERP: Maybe it's better to add purchase_price_delivery and store the standard price there.  Delivery margin calculations should then be based on this purchase price.

pedrobaeza: But you will have the same problem on multiple deliveries...

sergio-teruel: When you validate a picking (his stock moves) the product standard price is recomputed (denpending of valuation settings), this is the reason of this line.
And this field is added from sale_margin module, so i think that this module is the correct place...

chienandalu: > When you validate a picking (his stock moves) the product standard price is recomputed 

Then `_compute_margin` should be used to avoid currency and uom issues, don't you think?: https://github.com/odoo/odoo/blob/11.0/addons/sale_margin/models/sale_order.py#L14

Anyway, as stated by others, we'd still have the problem with multiple deliveries, backorders, etc. Maybe we could use  `product.price.history` and the mentioned `purchase_price_delivery` so we obtained an accurate margin?

sergio-teruel:
@chienandalu @pedrobaeza Can you review?

pedrobaeza (approved):
The new agreed method is much more flexible and allows both approaches.

PR #45, opened 2019-04-16
[11.0][IMP] sale_margin_delivered: take an account returned moves

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/account_invoice_margin/models/account_invoice.py
+++ b/account_invoice_margin/models/account_invoice.py
@@ -53,12 +53,6 @@ class AccountInvoiceLine(models.Model):
     def _onchange_product_id_account_invoice_margin(self):
         if self.invoice_id.type in ['out_invoice', 'out_refund']:
             purchase_price = self.product_id.standard_price
-            if any(self.product_id.supplier_taxes_id.mapped('price_include')):
-                taxes = self.product_id.supplier_taxes_id.compute_all(
-                    purchase_price, self.invoice_id.currency_id, 1,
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Why not using directly the quantity received shown in sale.order.line? It takes into account returned quantity when you mark the "To refund" check. With this, I think there will be inconsistent data.

sergio-teruel:
¿Quantity received?

pedrobaeza:
Sorry, "quantity delivered" is the term (I think the field is `qty_delivered`).

sergio-teruel:
Ok, i understand, I the customer return me goods but I do not check to refund the delivered qty is the initial delivered quantity.... and in this case i would discard this stock moves....

yajo (approved):

account_invoice_margin/models/account_invoice.py
```diff
@@ -53,12 +53,6 @@ def _compute_margin(self):
     def _onchange_product_id_account_invoice_margin(self):
         if self.invoice_id.type in ['out_invoice', 'out_refund']:
             purchase_price = self.product_id.standard_price
-            if any(self.product_id.supplier_taxes_id.mapped('price_include')):
```
pedrobaeza, line 56: Are you sure this is not needed?

sergio-teruel: Yes!! Also is incorrect do that!!

PR #48, opened 2019-07-30
[10.0] .travis.yml: Update to latest definition

FFernandez-PlanetaTIC:


pedrobaeza:
You can take the occasion to update whole template, removing one build and all the Transifex stuff.

FFernandez-PlanetaTIC:
I will try not to break anything...

FFernandez-PlanetaTIC:
`sale_line_cost_control` tests fail...

https://travis-ci.org/OCA/margin-analysis/jobs/565525241#L763
https://travis-ci.org/OCA/margin-analysis/jobs/565525242#L763

pedrobaeza:
This is something that happens because Odoo added an extra constraint when changing currency. What you can do it to set the currency by SQL in the tests as a workaround. It's what we have made in other repositories, but right now, I don't find any example.

PR #47, opened 2019-07-30
[10.0] [MIG] product_replenishment_cost

FFernandez-PlanetaTIC:

PR #49, opened 2019-07-30
[10.0] [MIG] product_standard_margin

FFernandez-PlanetaTIC:
Depends on #47


legalsylvain:
Hi @FFernandez-PlanetaTIC. thanks for this work. could you apply cherry-pick the history, like for the other PR ? 

Note : runbot should be green, as #47 is now merged.

thanks !

legalsylvain:
Also in the same way, I guess that rate should * 100. 
ref : https://github.com/OCA/sale-workflow/pull/909#issuecomment-517822595

what do you think ?

FFernandez-PlanetaTIC:
Ready for review

FFernandez-PlanetaTIC:
@legalsylvain,

There is only a commit in the history, this module was not removed on create branch 10.0.

Margin already was * 100.

Regards!

MPoch-PlanetaTIC:
I have added standard_margin and standard_margin_rate at product.template model. If product.template instance has only one variant it takes the standard_margin from its variant, otherwise standard_margin will be 0.

I renamed standard_price field as "Margin" and not "Theorical Margin".

legalsylvain:
Hi @FFernandez-PlanetaTIC 

> There is only a commit in the history, this module was not removed on create branch 10.0.

OK ! I did'nt saw. 

> I have added standard_margin and standard_margin_rate at product.template model. If product.template instance has only one variant it takes the standard_margin from its variant, otherwise standard_margin will be 0.

Good addition ! Thanks ! only one question, 0 is maybe not the best "not computable value" as this is a real margin. if margin is not computable, the current code use 999. value. What about to use it also for your case ? 

https://github.com/OCA/margin-analysis/pull/49/commits/eba1e1c9cb86c1cd64c21f9ddf36eee7942ff6de#diff-76d3fef1baafcb62886e4b80eb0c35d1L52


> I renamed standard_price field as "Margin" and not "Theorical Margin".

:+1:

MPoch-PlanetaTIC:
> Hi @FFernandez-PlanetaTIC
> 
> > There is only a commit in the history, this module was not removed on create branch 10.0.
> 
> OK ! I did'nt saw.
> 
> > I have added standard_margin and standard_margin_rate at product.template model. If product.template instance has only one variant it takes the standard_margin from its variant, otherwise standard_margin will be 0.
> 
> Good addition ! Thanks ! only one question, 0 is maybe not the best "not computable value" as this is a real margin. if margin is not computable, the current code use 999. value. What about to use it also for your case ?
> 
> [eba1e1c#diff-76d3fef1baafcb62886e4b80eb0c35d1L52](https://github.com/OCA/margin-analysis/commit/eba1e1c9cb86c1cd64c21f9ddf36eee7942ff6de#diff-76d3fef1baafcb62886e4b80eb0c35d1L52)
> 
> > I renamed standard_price field as "Margin" and not "Theorical Margin".
> 
> +1

Hi @legalsylvain 

I fixed the default value of standard_margin to 999. as you have suggested.

Thanks

legalsylvain (approved):
LGTM. Thanks !

PR #53, opened 2019-09-18
[10.0] [IMP] product_replenishment_cost: Add to product.template

MPoch-PlanetaTIC:
Adds replenishment_cost field to product.template model.
Instances of product.template with multiple product variants, have replenishment_cost 0.

cc @PlanetaTIC

### DIFF
--- a/product_replenishment_cost/models/product_product.py
+++ b/product_replenishment_cost/models/product_product.py
@@ -42,3 +42,29 @@ class ProductProduct(Model):
              "information, for example Bills of Materials or latest "
              "Purchases. By default, the Replenishment cost is the same "
              "as the Cost Price.")
+
+
+class ProductTemplate(Model):
+    _inherit = 'product.template'
+
+    replenishment_cost = fields.Float(
+        string='Replenishment cost', compute='_compute_replenishment_cost',
+        store=True, digits=dp.get_precision('Product Price'),
+        help="The cost that you have to support in order to produce or "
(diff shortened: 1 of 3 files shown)


product_replenishment_cost/models/product_product.py
```diff
@@ -42,3 +42,29 @@ def _get_replenishment_cost(self):
...
+    @api.depends('product_variant_ids',
+                 'product_variant_ids.replenishment_cost')
+    def _compute_replenishment_cost(self):
+        unique_variants = self.filtered(
+            lambda template: len(template.product_variant_ids) == 1)
+        for template in unique_variants:
+            template.replenishment_cost = \
+                template.product_variant_ids.replenishment_cost
+        for template in (self - unique_variants):
+            template.replenishment_cost = 0.0
```
legalsylvain, line 70: if they are 3 variants with the same replenishment cost, it should be the value in the template, don't you think ?

legalsylvain (changes requested):
Hi @MPoch-PlanetaTIC. 

thanks for this improvment.

one question. LGTM otherwise.

regards.

product_replenishment_cost/models/product_product.py
MPoch-PlanetaTIC: I did this way because standard_price is set the same way: Zero if the template has multiple product variants.
It can be confuse for users if some templates shows the replenishment_cost and some others not.
It's true, the way I have done can also be confusing, but standard_price works the same way.

PR #54, opened 2019-09-20
[10.0] [ADD] add product_replenishment_cost_default_seller

MPoch-PlanetaTIC:
This new module compute's replenishment_cost field using default supplierinfo's price. If a product instance has no supplierinfo (seller_ids field is empty) then super is called to compute replenishment_cost.

cc @PlanetaTIC


product_replenishment_cost_default_seller/models/product_product.py
```diff
@@ -0,0 +1,31 @@
...
+    @api.one
+    @api.depends('product_tmpl_id.standard_price', 'standard_price',
+                 'product_tmpl_id.seller_ids.price')
+    def _get_replenishment_cost(self):
+        if self.seller_ids:
+            self.replenishment_cost = self.seller_ids[0].price
+        else:
+            super(ProductProduct, self)._get_replenishment_cost()
+
+    replenishment_cost = fields.Float(
```
legalsylvain, line 12: why do you redefine this field ?

product_replenishment_cost_default_seller/models/product_product.py
```diff
@@ -0,0 +1,31 @@
...
+from odoo import fields, api
+from odoo.models import Model
+
+import odoo.addons.decimal_precision as dp
+
+
+class ProductProduct(Model):
+    _inherit = 'product.product'
+
+    @api.one
```
legalsylvain: it's an api.multi. 
api.one is deprecating.

MPoch-PlanetaTIC: Fixed.
Thanks for your feedback

product_replenishment_cost_default_seller/models/product_product.py
MPoch-PlanetaTIC: Was thinking on old OpenERP versions.

Fixed.

PR #55, opened 2019-09-23
10.0 add product replenishment cost supplier discount

MPoch-PlanetaTIC:
This new module gets installed when product_replenishment_cost_default_seller and product_supplierinfo_discount are installed.

With this module, replenishment_cost is computed using supplierinfo's price and supplierinfo's discount. If a product instance has no supplierinfo (seller_ids field is empty) then super is called to compute replenishment_cost.

cc: @PlanetaTIC


legalsylvain:
Hi. Interesting module. Did you planned to make similar module for triple_discount module ?
Kind regards.

PR #62, opened 2019-11-29
[MIG][12.0] sale_report_margin: Migration to v12.0

sergio-teruel:
cc @Tecnativa TT20187
@chienandalu @rafaelbn Can you review?


chienandalu (approved):
Looks good :+1:

sale_report_margin/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Sale Report Margin",
+    "version": "12.0.1.0.0",
+    "author": "Tecnativa,"
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-reporting",
+    "development_status": "Production/Stable",
```
pedrobaeza, line 9: You need to declare `maintainers` if promoting this to stable

sale_report_margin/reports/sale_report.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SaleReport(models.Model):
+    _inherit = "sale.report"
+
+    purchase_price = fields.Float(
+        string='Purchase Price',
+        readonly=True,
+    )
+
+    def _select(self):
```
pedrobaeza: This has been modified upstream. The hint about this is the coverage drop. Now everything goes in one method:

https://github.com/odoo/odoo/blob/a7d24cabafe1a9c71f9dbdb7cf3f129006250951/addons/sale/report/sale_report.py#L58

and in this case it's simple, as you only have to add the field in the argument:

```python
def _query(self, with_clause='', fields={}, groupby='', from_clause=''):
    fields['purchase_price'] = ',\nSUM(l.purchase_price / CASE COALESCE(s.currency_rate, 0)) as purchase_price'
    return super()._query(self, with_clause=with_clause, fields=fields, groupby=groupby, from_clause=from_clause)
```
My only doubt (but it's an Odoo problem) is if the argument `field={}` will have problem in a multi-DB (https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python).

pedrobaeza (changes requested):

sergio-teruel:
Changes done!!

pedrobaeza:
Everything is red

sergio-teruel:
Changes done

pedrobaeza (approved):

PR #64, opened 2019-12-03
[11.0][FIX] account_invoice_margin: Set purchase_price when user has not se invoice margin security group set

### LINKED ISSUE #52: Bug in account_invoice_margin
If you create an invoice without the **Show Invoice Margin** security group, the `purchase_price` field of `account.invoice.line` model isn't set.

**Steps to reproduce**:
1. With a user who doesn't have **Show Invoice Margin** permission
2. Create an invoice
3. Give permissions to the user to see the margins (Show Invoice Margin)
4. Press `ctrl+r` to reload
5. Check the invoice line. The margin percentage will be 100% and the cost will be 0.

imagen

Can it be related to the way the view is restricted to the group?

https://github.com/OCA/margin-analysis/blob/366c90b2f54901f4aff25098fb6a011f2adc4b2a/account_invoice_margin/views/account_invoice_margin_view.xml#L24-L25

sergio-teruel:
cc @Tecnativa
Fixes #52

### DIFF
--- a/account_invoice_margin/views/account_invoice_margin_view.xml
+++ b/account_invoice_margin/views/account_invoice_margin_view.xml
@@ -2,48 +2,84 @@
 <!-- © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3). -->
 <odoo>
+    <!--
+    The xxx_to_save views only adds purchase price field in view to allow
+    store this value because this field must be in view.
+    Other views display this field and others for users if they have the
+    security group.
+    -->
+    <record id="invoice_margin_form_tree_to_save" model="ir.ui.view">
+        <field name="model">account.invoice</field>
+        <field name="inherit_id" ref="account.invoice_form"/>
+        <field name="arch" type="xml">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As the field is not a many2one to other forbidden model, I think it's more convenient to have the field as invisible in main view, and have a new view only for the margin group that set the invisibility to 0, and thus you don't need any extra code.

sergio-teruel:
Changes Done

account_invoice_margin/views/account_invoice_margin_view.xml
```diff
@@ -2,48 +2,84 @@
 <!-- © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3). -->
 <odoo>
+    <!--
+    The xxx_to_store views only adds purchase price field in view to allow
```
pedrobaeza:
```suggestion
    The xxx_to_save views only adds purchase price field in view to allow
```

pedrobaeza (approved):

sergio-teruel:
Done!!

carlosdauden (approved):

PR #68, opened 2019-12-11
[IMP][12.0] account_invoice_margin

legalsylvain:
[IMP] make purchase price computation overloadable [Reference](https://github.com/OCA/margin-analysis/pull/66#issuecomment-564623859).
[REF] rename file with the name of the model
[ADD] margins on account.invoice model (and related views). (Restore functionnality present in the OCA module [invoice_margin](https://github.com/OCA/account-invoicing/tree/8.0/invoice_margin/models) in V8)
[ADD] description and screenshot
[FIX] Typo on margin (%) field name
[ADD] fr translation

account_invoice_form

account_invoice_form_tree


account_invoice_margin/models/account_invoice.py
```diff
@@ -5,6 +5,54 @@
...
+    )
+
+    @api.multi
+    @api.depends(
+        'invoice_line_ids.margin',
+        'invoice_line_ids.margin_signed',
+        'invoice_line_ids.price_subtotal',
+    )
+    def _compute_margin(self):
+        for invoice in self:
```
pedrobaeza: Please do this through `read_group` for performance.

account_invoice_margin/readme/DESCRIPTION.rst
```diff
@@ -1 +1,11 @@
 This module allows to show sales margin in invoices.
+
+The new informations are available :
```
pedrobaeza:
```suggestion
The new information is available: 
```

legalsylvain: thanks !

account_invoice_margin/models/account_invoice.py
legalsylvain: Hi @pedrobaeza. thanks for your review.

sorry, I'm not sure to understand. Do you want me to write something like : 

    AccountInvoiceLine.read_group(
        [('invoice_id', 'in', self.ids)],
        ['margin', 'margin_signed', 'price_subtotal'],
        'invoice_id'
    )

pedrobaeza: Yes, that's it, and you get directly the sum. Take care that read_group syntax is not exactly that, but you got the point.

legalsylvain: Hi I tried this code but, the onchange is so failing. (when I change a line, the invoice margin is set to 0)
Do you know how to fix it ? Or did I missed something ?

        for invoice in self:
            res = self.env['account.invoice.line'].read_group(
                [('invoice_id', '=', self.id)],
                ['margin', 'margin_signed', 'price_subtotal'],
                ['invoice_id'],
            )
            if res:
                res = res[0]
                invoice.update({
                    'margin': res['margin'],
                    'margin_signed': res['margin_signed'],
                    'margin_percent': res['price_subtotal'] and (
                        res['margin_signed'] / res['price_subtotal'] * 100),
                })

The problem I see with what you propose is that self.ids (or self.id) doesn't have allways a value.

pedrobaeza: You have to use `res.get('field', 0)`

legalsylvain: That's not the problem. The problem is : 

- Open a draft invoice
- change a value in a line
- the function is called but  ``res == []``
-> the values are set to 0 -> KO

This behaviour does'nt occure with the current code of the PR.

pedrobaeza: You can add a special case for NewId, but the performance gets really improved with this for the 99% percentage of cases (stored data).

legalsylvain: OK ! It works ! 

Better ? : https://github.com/OCA/margin-analysis/pull/68/files#diff-53256610b917670c26058816b8fcaac1R41

pedrobaeza: Not really:

- you are still losing the performance doing N `read_group` calls: do only once with all the real IDs together.
- You are not computing for the virtual IDs via the normal computation (`sum('invoice_line_ids.margin...)`).

pedrobaeza: Do you want me to do a PR over your branch with this?

account_invoice_margin/demo/res_groups.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<odoo>
+
+    <record id="group_account_invoice_margin_security" model="res.groups">
+        <field name="users" eval="[(4, ref('base.user_admin'))]"/>
```
pedrobaeza: Isn't better to put this as regular data and not demo one?

account_invoice_margin/models/account_invoice.py
legalsylvain: Yes, please. Sorry for not understanding.
I made some grep in other repo, but did not found read_group in other compute method, to have inspiration.

pedrobaeza:
@legalsylvain I have pushed the code with all the techniques for performance and without drawbacks.

account_invoice_margin/demo/res_groups.xml
legalsylvain: done.

legalsylvain:
Hi @pedrobaeza 
First, happy new year ! 
1) I updated moving res_group file in data folder, according your remark.
2) tested your changes. LGTM.

could you update your review  ? 

Thanks !

pedrobaeza (approved):
Happy New Year to you too!

Thanks for the double checking

legalsylvain:
@sergio-teruel : could you take a look, as you are set as maintainer ?

thanks !

sergio-teruel (approved):

PR #76, opened 2020-03-21
[MIG][12.0] product_margin_classification (from OCA/sale-workflow/10.0)

legalsylvain:
migration of the module product_margin_classification (previously in sale-worklow for the [8.0](https://github.com/OCA/sale-workflow/pull/316) and the [10.0](https://github.com/OCA/sale-workflow/pull/909) branches.)

- classic migration.
- Removed obsolete needaction inheritance.
- improved classification form view, adding smart buttons

margin_classification_form

CC : Original reviewers : @quentinDupont, @MPoch-PlanetaTIC, @FFernandez-PlanetaTIC, @gurneyalex

Note : @gurneyalex, I took [your advice](https://github.com/OCA/sale-workflow/pull/316#issuecomment-241780696), and removed the icon module.
Hope you'll appreciate. ;-)


MPoch-PlanetaTIC:
Hello @legalsylvain ,

I was testing your migration and buttons "Increase Price" and "Descrease Price" from product.product view fail when clicking from a product variant (product.product), they work properly when using from product.template.

It seems you need to create function use_theoretical_price on product.product model.

Regards

MPoch-PlanetaTIC:
> Hello @legalsylvain ,
> 
> I was testing your migration and buttons "Increase Price" and "Descrease Price" from product.product view fail when clicking from a product variant (product.product), they work properly when using from product.template.
> 
> It seems you need to create function use_theoretical_price on product.product model.
> 
> Regards

I have read the Roadmap now.

I don't know the best solution, but currently an internal Odoo error is raised directly to the user, maybe should be better to raise an error message saying that margin classification only works on product templates?

Regards

legalsylvain:
> I was testing your migration and buttons "Increase Price" and "Descrease Price" from product.product view fail when clicking from a product variant (product.product), they work properly when using from product.template.

Thanks, I'll fix that.

> It seems you need to create function use_theoretical_price on product.product model.

Indeed !

thanks for your review.

legalsylvain:
hi @MPoch-PlanetaTIC. I fixed the code. Could you test again ? Thanks !

MPoch-PlanetaTIC:
> hi @MPoch-PlanetaTIC. I fixed the code. Could you test again ? Thanks !

Hello @legalsylvain 
I have tested your fix and it works properly.

But I would like to notice you a problem we found using this module in Odoo 10.

We had performance problems on an Odoo 10 instance with thousands of products and with purchase orders with hundreds of lines:
 When receiving purchase orders, every product reception updates standard_price and because field dependency on standard_price, module product_margin_classification recomputes theoretical_price, theoretical_difference, margin_state and updating theoretical_difference.
 Updating theoretical_difference makes to recompute the product.margin.classification's fields: template_correct_price_qty, template_incorrect_price_qty, template_too_cheap_qty and template_too_expensive_qty.
 Recomputing these fields took long time when using an Odoo database with thousands of products. And because purchase orders contains lot of lines, it took even longer.

And this time consummation took us to a second problem: Database concurrency problems if user edited standard_price (manual editing or some other process updating standard_price).

To solve these problems we made _compute_theoretical_multi to depends not on standard_price but on replenishment_cost (field introduced in product_replenishment_cost module of margin-analysis repository). And in product_replenishment_cost module, made replenishment_cost field to depends on seller_ids's price and not standard_price.

Just wanted to inform you about this issue, maybe you are interested to adapt this module to the situation we found.

Rewards

FFernandez-PlanetaTIC:
Hi @legalsylvain ,

I just want to add that:
- Performance and concurrency issues were **very heavies**. Maybe we could warn about these problems in the README. As @MPoch-PlanetaTIC said, it happens when cost method is average/last, with thousands of products, hundreds of purchase lines and margin should be computed from cost price.
- We made _product_margin_classification_ to depend on _product_replenishment_cost_, for other reasons, in customer customizations. Maybe  the change could be done in the module itself.
- We changed the behavior of _product_replenishment_cost_ with these new modules:
  - _product_replenishment_cost_default_seller_ OCA/margin-analysis#54
  - _product_replenishment_cost_supplier_discount_ OCA/margin-analysis#55

Regards

legalsylvain:
Hi @MPoch-PlanetaTIC 
thanks for your review. 
I think that adding a dependency to product_replenishment_cost is a good idea.

> To solve these problems we made _compute_theoretical_multi to depends not on standard_price but on replenishment_cost (field introduced in product_replenishment_cost module of margin-analysis repository).

I'd like to know how you implemented that. Did you moved all the "margin code" from template to product ? 
kind regards.

legalsylvain:
@MPoch-PlanetaTIC
Could you point the code where you made the dependency of product_replenishment_cost.

thanks !

MPoch-PlanetaTIC:
> @MPoch-PlanetaTIC
> Could you point the code where you made the dependency of product_replenishment_cost.
> 
> thanks !

Hello @legalsylvain sorry for the long silence,

In short, we are suggesting you to edit lines 49 and 77 of file models/product_template.py replacing "standard_price" by "replenishment_cost" and adding the dependency to product_replenishment_cost module.

But we solved this problem creating a new custom module and overriding the fields computed by _compute_theoretical_multi function and the function itself.

houssine78:
Hi @legalsylvain 

Do you think you will address those changes ? if yes we'll wait a bit before putting this module in production.

Thank you for your feedback !

Houssine

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+    @api.model
+    def _default_price_round(self):
+        decimal_obj = self.env["decimal.precision"]
+        return 10 ** (-decimal_obj.precision_get("Product Price"))
+
+    # constrains Section
+    @api.constrains("markup")
+    def _check_markup(self):
+        for classification in self:
+            if classification.markup == -100:
```
rousseldenis: You need to use float_compare as using digit precision

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+                raise UserError(_("-100 is not a valid Markup."))
```
rousseldenis: A constrains must raise a ValidationError instead

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+
+    @api.multi
+    def _inverse_profit_margin(self):
+        pass
+
+    @api.onchange("profit_margin")
+    def _onchange_profit_margin(self):
+        for classification in self:
+            if classification.profit_margin == 100:
```
rousseldenis: Same as above

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+                classification.profit_margin
+                / (100 - classification.profit_margin)
+            )
+
+    # Compute Section
+    @api.multi
+    @api.depends("markup")
+    def _compute_profit_margin(self):
+        for classification in self:
+            if classification.markup != -100:
```
rousseldenis: Same as above.

Moreover, you need to initialize all records in self with a default value.

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+                classification.profit_margin = (
+                    1 - (1 / (classification.markup / 100 + 1))
+                ) * 100
+
+    @api.multi
+    @api.depends("template_ids.theoretical_difference")
+    def _compute_template_different_price_qty(self):
+        for classification in self:
+            template_vals = classification.template_ids.read(["margin_state"])
```
rousseldenis: You MUST depends on all fields for stored values. So, template_ids.margin_state.

With this implementation, you do an SQL query at each loop... So, get confident in Odoo ORM and use the good depends.

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+            classification.template_too_cheap_qty = len([
```
rousseldenis: So, after having changed the depends, you can use len(classification.template_ids.filtered(lambda t: t.margin_state == "too_cheap"))

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+                x["id"] for x in template_vals
+                if x["margin_state"] == "correct"
+            ])
+            classification.template_incorrect_price_qty = (
+                classification.template_too_expensive_qty
+                + classification.template_too_cheap_qty
+            )
+
+    @api.multi
+    @api.depends("template_ids.margin_classification_id")
```
rousseldenis: Why not just 'template_ids' ?

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+        for classification in self:
+            classification.template_qty = len(classification.template_ids)
+
+    # Constraint Section
+    @api.multi
+    @api.constrains("price_round")
+    def _check_price_round(self):
+        for classification in self:
+            if classification.price_round == 0:
+                raise UserError(_("Price Rounding can not be null."))
```
rousseldenis: Same as above, must raise ValidationError

product_margin_classification/models/product_margin_classification.py
```diff
@@ -0,0 +1,188 @@
...
+    def _compute_template_qty(self):
+        for classification in self:
+            classification.template_qty = len(classification.template_ids)
+
+    # Constraint Section
+    @api.multi
+    @api.constrains("price_round")
+    def _check_price_round(self):
+        for classification in self:
+            if classification.price_round == 0:
```
rousseldenis: Same as above (float_is_zero here)

product_margin_classification/models/product_template.py
```diff
@@ -0,0 +1,99 @@
...
+                    tools.float_round(
+                        template.standard_price * multi,
+                        precision_rounding=classification.price_round,
+                    )
+                    + classification.price_surcharge
+                )
+            else:
+                template.theoretical_price = template.list_price
+            difference = template.list_price - template.theoretical_price
+            if max(difference, -difference) < 10 ** -9:
```
rousseldenis: Why this arbitrary value ? I think this is due to no use of float_compare. Use it in every comparison here under and remove this max test.

rousseldenis (changes requested):
Code review.

The use of float_compare essentially and performance part.

A little bit more tests should be great

product_margin_classification/models/product_margin_classification.py
legalsylvain: Done.

product_margin_classification/models/product_margin_classification.py
legalsylvain: Done.

product_margin_classification/models/product_margin_classification.py
legalsylvain: Done.

product_margin_classification/models/product_margin_classification.py
legalsylvain: no need to initialize now, as I raise an error if markup == 100. (inconsistent setting)

product_margin_classification/models/product_margin_classification.py
legalsylvain: Solved. I set store to False.

product_margin_classification/models/product_margin_classification.py
legalsylvain: now store = False, so i think it's oK.

product_margin_classification/models/product_margin_classification.py
legalsylvain: done.

product_margin_classification/models/product_margin_classification.py
legalsylvain: done.

product_margin_classification/models/product_margin_classification.py
legalsylvain: done.

product_margin_classification/models/product_template.py
legalsylvain: ok, I used float_compare also. more simple BTW ! thanks !

legalsylvain:
- just set fields as store = False, and computed on the fly, on the model ``product.margin.classification``. This should reduce the performance issue @MPoch-PlanetaTIC could face.
- move fields from template to product model, with according migration script, for consistency reasons.
- various fixes, regarding @rousseldenis remarks. Thanks a lot for your deep review ! I didn't use float_compare.
- also improved a little coverage.

thanks a lot for all your reviews.

legalsylvain:
Hi @rousseldenis : could you take a look ? I adressed your questions; kind regards.

rousseldenis:
Oops, didn't see but you need to keep commit history even when changing repo.

legalsylvain:
HI @rousseldenis 
I didn't know it was possible. Could you point how to do it ? I'm not git expert, and only know the oca convention, in the same repo.
ref : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

Thanks.

rousseldenis:
> HI @rousseldenis
> I didn't know it was possible. Could you point how to do it ? I'm not git expert, and only know the oca convention, in the same repo.
> ref : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0
> 
> Thanks.

Yeah, that's a little bit more tricky.

You can add a remote to sale-workflow, then when executing git format-patch command, point out to "sale-workflow remote"/10.0. That should do the job

legalsylvain:
indeed. That worked ! Thanks.
could you update your review ? 

```
git remote add sale-workflow https://github.com/OCA/sale-workflow
git fetch sale-workflow 
git format-patch --keep-subject --stdout OCA/12.0..sale-workflow/10.0 -- product_margin_classification | git am -3 --keep
```

rousseldenis:
Seems OCB is green but not Odoo. Could you check travis?

quentinDupont (approved):

quentinDupont:
Tested on runbot and seems all good :clap: 
Travis and runbot are good :green_circle:

marcelsavegnago (approved):
LGTM

legalsylvain:
@rousseldenis It's seems all is OK now., if yes, could you unblock the PR ?
thanks !

leonhandreke:
Hi, sorry to be pushy here. It looks to me like this was approved but somehow wasn't because the right person wasn't around. We are using this module in production, so it would be really nice if this was merged at some point. Please let me know if there is anything I can do here.

houssine78:
Hi @pedrobaeza, Denis hasn't the permission to merge. maybe you could help on getting this pr merged ? Thanks

pedrobaeza:
A rebase is needed and then check the problem

houssine78:
thanks @pedrobaeza !

@legalsylvain could you rebase and then ping pedro once it's done ?

dreispt:
Travis is red

legalsylvain:
Travis is green. any maintainers could merge this approved PR ? 
thanks.

legalsylvain:
thanks @pedrobaeza !

PR #77, opened 2020-06-01
[FIX] sale_margin_security: onchange product

chienandalu:
In sale_margin, when the line already exists, the purchase_price
is computed with the product onchange method, so if the user doesn't
have the field available in the view it's not going to get the right
price. This should be fixed in v13 with the new compute fields
possibilities.

@Tecnativa TT24134

### DIFF
--- a/sale_margin_security/__init__.py
+++ b/sale_margin_security/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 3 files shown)


sergio-teruel:
@chienandalu Can you add some test?

rafaelbn (approved):
:+1: 

Add test please

pedrobaeza:
Cherry-picked for v12 in 83c3309ae41bd9058556320fc60fafe507099127 and for v13 in 0f764843a47de5a21eb1242486727435c25e83ec. The computed writable possibility hasn't been done yet.

PR #81, opened 2020-09-18
[FIX] sale_margin_delivered: Fix compute avco from stock valuation layer instead of stock move

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/sale_margin_delivered/__manifest__.py
+++ b/sale_margin_delivered/__manifest__.py
@@ -7,7 +7,7 @@
     "website": "https://github.com/OCA/margin-analysis",
     "category": "Sales",
     "license": "AGPL-3",
-    "depends": ["stock", "sale_margin_security"],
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why `move.price_unit` is not reliable?

sergio-teruel:
Odoo V13.0 for outgoing moves from sales does not store this field. Now is stored into stock valuation layer model.
So price_unit stock_move field for sales moves always is zero..

sergio-teruel:
I will explainit into code moule...

pedrobaeza:
Thanks for the explanation. In this case, I would say to put that explanation in the commit message instead of the code, but in code can also serve for explanation. As you prefer.

pedrobaeza (approved):

marcelsavegnago (approved):

PR #84, opened 2020-10-14
[FIX] stored field should not depend on none stored field

sebastienbeau:
if a stored field depend on a non stored field, really weird behaviour
can occure.
Indeed none-stored field can not be searched correctly and the ORM will
return all product (with an empty domain).
So in my case I import data through csv field one product.template.attribute.value
it will recompute the standard price on all product of my database !

### DIFF
--- a/product_standard_margin/models/product_product.py
+++ b/product_standard_margin/models/product_product.py
@@ -43,8 +43,8 @@ class ProductProduct(models.Model):
 
     # Compute Section
     @api.depends(
-        "lst_price",
         "product_tmpl_id.list_price",
+        "product_tmpl_id.price_attribute_value_ids.price_extra",
         "standard_price",
         "taxes_id.price_include",
         "taxes_id.amount",
--- a/product_standard_margin/models/product_template.py
+++ b/product_standard_margin/models/product_template.py
@@ -37,10 +37,14 @@ class ProductTemplate(models.Model):
         "Take care of tax include and exclude.. If no sale price "
         "set, will display 999.0",
     )
+    price_attribute_value_ids = fields.One2many(
+        "product.template.attribute.value",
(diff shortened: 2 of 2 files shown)


sebastienbeau:
@mourad-ehm can you check this

product_standard_margin/models/product_template.py
```diff
@@ -37,10 +37,14 @@ class ProductTemplate(models.Model):
         "Take care of tax include and exclude.. If no sale price "
         "set, will display 999.0",
     )
+    price_attribute_value_ids = fields.One2many(
```
legalsylvain, line 40: Hi. Could you mention where this field is initialized ?

product_standard_margin/models/product_product.py
```diff
@@ -43,8 +43,8 @@ class ProductProduct(models.Model):
 
     # Compute Section
     @api.depends(
-        "lst_price",
         "product_tmpl_id.list_price",
+        "product_tmpl_id.price_attribute_value_ids.price_extra",
```
legalsylvain, line 47: this field doesn't change the computation of the margin. why did you add it ?

product_standard_margin/models/product_template.py
```diff
@@ -37,10 +37,14 @@ class ProductTemplate(models.Model):
...
+        "product.template.attribute.value",
+        "product_tmpl_id",
+        )
 
     # Compute Section
     @api.depends(
-        "lst_price",
+        "list_price",
```
legalsylvain, line 47: indeed. Good catch !

legalsylvain (changes requested):
Hi Seb. Thanks for your work.
i don't get all the changes. remarks inline.

legalsylvain:
ping @sebastienbeau

PR #87, opened 2020-12-22
14.0 mig sale order margin percent

Kev-Roche:


legalsylvain (approved):
Diff review. LGTM. Thanks.

mourad-ehm:
Hi @Kev-Roche  @legalsylvain , 
I added percent on sale lines.
in 12.0 here
https://github.com/OCA/margin-analysis/pull/88

legalsylvain (commented):
As said in #88 

I think 
- ``percent`` field should be renamed to ``margin_percent`` field. (as in ``account_invoice_margin`` module, same repository)
- percent fields should be stored, to avoid to recompute each time.
- adding a button in a tree view just to diplay a text is not in the Odoo spirit IMO.

otherwise, thanks for porting this module !

mourad-ehm:
Thanks @legalsylvain for your review. 
Ok for remark. 
but do you have some solution to display text in tree view ?

legalsylvain:
>  but do you have some solution to display text in tree view ?

not for the time being, but this is like the other module.

I think that if it's a real need, it could be done by a widget. (like ``widget=percent``), to do.

mourad-ehm:
Ok. Thanks.
The best widget is which can add suffix or prefix text to field value.
I will remove button for now while waiting for the widget to be ready

mourad-ehm:
This module is no longer necessary.
 it is integrated in odoo sale_margin module  https://github.com/odoo/odoo/blob/14.0/addons/sale_margin/models/sale_order.py#L12

PR #88, opened 2020-12-24
[IMP] add margin percent on order_line

mourad-ehm:


sale_order_margin_percent/models/sale_order_margin.py
```diff
@@ -17,3 +17,19 @@ def _compute_percent(self):
...
+
+    percent = fields.Float(
+        string='Percent',
+        compute='_compute_percent',
+        digits=(16, 2),
+        )
+
+    @api.depends('margin', 'price_subtotal')
+    def _compute_percent(self):
+        for order in self:
```
legalsylvain, line 35:
```
for order_line
    order_line.margin_percent = (order_line.margin / order_line.price_subtotal) * 100 if order_line.price_subtotal else 0.0
```

see : https://github.com/OCA/margin-analysis/blob/12.0/account_invoice_margin/models/account_invoice.py#L119

sale_order_margin_percent/views/sale_order_margin_percent_view.xml
```diff
@@ -8,6 +8,16 @@
             <field name="margin" position="after">
                 <field name="percent" widget="progressbar"/>
             </field>
+            <xpath expr="//field[@name='order_line']/form//field[@name='tax_id']" position="before">
+                <field name="percent" string="Margin" widget="progressbar"/>
+            </xpath>
+           <xpath expr="//field[@name='order_line']/tree//field[@name='price_subtotal']" position="after">
+                <field name="percent" string="Margin"/>
```
legalsylvain: you can remove ``string=`` as the default string is oK.

sale_order_margin_percent/models/sale_order_margin.py
```diff
@@ -17,3 +17,19 @@ def _compute_percent(self):
...
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    percent = fields.Float(
+        string='Percent',
+        compute='_compute_percent',
+        digits=(16, 2),
+        )
```
legalsylvain, line 31:
```
    margin_percent = fields.Float(
        string='Margin (%)',
        compute='_compute_percent',
        digits=dp.get_precision('Product Price'),
        )
```

to be consistent with accout_invoice_margin module. (same repository)
https://github.com/OCA/margin-analysis/blob/12.0/account_invoice_margin/models/account_invoice.py#L95

sale_order_margin_percent/views/sale_order_margin_percent_view.xml
```diff
@@ -8,6 +8,16 @@
...
+                <!-- The button is used only to display % icon
+                    since is not possible to add html to tree view yet-->
+                <button icon="fa-percent"/>
```
legalsylvain: not sure this hack (adding a button just for a label, AFAIK) is interesting.

See invoice module : https://github.com/OCA/margin-analysis/blob/12.0/account_invoice_margin/views/account_invoice_margin_view.xml#L84

legalsylvain (changes requested):
Hi @mourad-ehm. thanks for the improvement.

some comments inline, to have the same kind of module as ``account_invoice_margin``.

One question, do you know why the ``margin_percent`` field is not stored. (it is stored in ``account_invoice_margin`` and I see it better, because otherwise, it require useless recompute in each display of the sale order).

kind regards.

PR #72, opened 2020-02-22
[MIG][13.0] account_invoice_margin: Migration to v13.0

sergio-teruel:
cc @Tecnativa TT21603

PR #89, opened 2021-02-01
[FIX] account_invoice_margin, account_invoice_margin_sale: add generice method for better inheritance

bodedra:


bodedra:
@sergio-teruel @pedrobaeza fast-track?

sergio-teruel:
Sorry @bodedra but account_invoice_margin in v13.0 has not sale dependency...
Are you sure that you has lastest code??

https://github.com/OCA/margin-analysis/blob/13.0/account_invoice_margin/__manifest__.py#L16

bodedra:
Yes. Correct. That's why I proposed this PR to add `sale` dependency.

`sale_line_ids` declared in sale module and we are used field at https://github.com/OCA/margin-analysis/blob/13.0/account_invoice_margin/models/account_invoice.py#L44

So it breaks my unit test with `AttributeError: 'account.move.line' object has no attribute 'sale_line_ids'`

pedrobaeza:
@sergio-teruel you added such hidden dependency in #72

sergio-teruel:
Yes... A mistake!!
I am thinking that the better solution is move this condition to account_invoice_margin_sale.
Mybe you can add a method to filter records and in account_invoice_margin_sale module do a filter by downpayment..

What are you think about this?

bodedra:
We can not call super method in account_invoice_margin_sale. We have to rewrite entire method, without super. Will it work for you?

pedrobaeza:
A possible solution is:
```python
    def _get_margin_applicable_lines(self):
        self.ensure_one()
        return self.invoice_line_ids

    @api.depends(
        "invoice_line_ids.margin",
        "invoice_line_ids.margin_signed",
        "invoice_line_ids.price_subtotal",
    )
    def _compute_margin(self):
        for invoice in self:
            margin = 0.0
            margin_signed = 0.0
            price_subtotal = 0.0
            for line in invoice._get_margin_applicable_lines():
                margin += line.margin
                margin_signed += line.margin_signed
                price_subtotal += line.price_subtotal
            invoice.margin = margin
            invoice.margin_signed = margin_signed
            invoice.margin_percent = (
                price_subtotal and margin_signed / price_subtotal * 100 or 0.0
            )
```
and in `account_invoice_margin_sale`:
```python
    def _get_margin_applicable_lines(self):
        lines = super()._get_margin_applicable_lines()
        return lines.filtered(lambda x: not x.is_downpayment)
```

account_invoice_margin/models/account_invoice.py
```diff
@@ -40,7 +44,7 @@ def _compute_margin(self):
             margin = 0.0
             margin_signed = 0.0
             price_subtotal = 0.0
-            for line in invoice.invoice_line_ids:
+            for line in invoice._get_margin_applicable_lines():
                 if any(line.sale_line_ids.mapped("is_downpayment")):
```
pedrobaeza: You should remove this and if you can do the other change as well in the same PR.

bodedra: Yes. I am on it.

pedrobaeza:
What about `account_invoice_margin_sale`?

bodedra:
@pedrobaeza Would you like two PR or one PR with two module change?

pedrobaeza:
I think the change can be included in the same PR as it's very related.

bodedra:
@pedrobaeza @sergio-teruel It is ready for your review.

pedrobaeza (approved):
Thanks for discovering the problem

bodedra:
@pedrobaeza @sergio-teruel Appreciate your support. This PR is :green_apple:

bodedra:
We need to same thing for account.move.line object. https://github.com/OCA/margin-analysis/blob/92a40ed7afbcbfb85f18ef4a8dcfb5a86e5dec8a/account_invoice_margin/models/account_invoice.py#L87

pedrobaeza:
Yes, in this case is to extract this to `account_invoice_margin_sale`:
```python
    @api.depends("purchase_price", "price_subtotal")
    def _compute_margin(self):
        for line in self:
            if (
                line.move_id
                and line.move_id.type[:2] == "in"
                or any(line.sale_line_ids.mapped("is_downpayment"))
            ):
                line.update(
                    {"margin": 0.0, "margin_signed": 0.0, "margin_percent": 0.0}
                )
            else:
                super(AccountMoveLine, line)._compute_margin()
```

sergio-teruel (approved):

sergio-teruel:
Thanks @bodedra

marcelsavegnago (approved):

account_invoice_margin_sale/models/account_invoice.py
```diff
@@ -0,0 +1,30 @@
...
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.depends("purchase_price", "price_subtotal")
+    def _compute_margin(self):
+        for line in self:
+            if (
+                line.move_id
+                and line.move_id.type[:2] == "in"
+                or any(line.sale_line_ids.mapped("is_downpayment"))
```
pedrobaeza: I didn't see that the other condition is met in `super`, so no need to add it here:
```suggestion
                any(line.sale_line_ids.mapped("is_downpayment"))
```

bodedra: Addressed!

pedrobaeza (approved):
Thank you very much!

PR #90, opened 2021-03-09
[14.0][MIG] account_invoice_margin: Migration to 14.0

Reyes4711-S73:
Standard migration to 14.0


dreispt:
Testes are failing https://travis-ci.com/github/OCA/margin-analysis/jobs/491125211#L1128

pedrobaeza:
Please include #93

Reyes4711-S73:
Included @pedrobaeza

account_invoice_margin/__manifest__.py
```diff
@@ -13,7 +13,7 @@
     "maintainers": ["sergio-teruel"],
     "application": False,
     "installable": True,
-    "depends": ["account"],
+    "depends": ["account", "sale"],
```
pedrobaeza: Why adding this dependency?

account_invoice_margin/tests/test_account_invoice_margin.py
```diff
@@ -143,7 +142,7 @@ def test_invoice_down_payment(self):
         action = wiz.create_invoices()
         invoice_id = action["res_id"]
         invoice1 = AccountMove.browse(invoice_id)
-        self.assertEqual(invoice1.margin, 0.0)
+        self.assertEqual(invoice1.margin, 100.0)
```
pedrobaeza: Why the tests changes its results?

Reyes4711-S73: Because they are showing another value and I tested the module and the calculus are working fine

pedrobaeza: Which other value? It should displays the same value...

account_invoice_margin/__manifest__.py
Reyes4711-S73: The test needs this dependency

pedrobaeza: Then there's something incorrect. There's the module `account_invoice_margin_sale` that makes the link.

account_invoice_margin/tests/test_account_invoice_margin.py
Reyes4711-S73: I review again the calculus in the test and in a clean db and it's working fine. The change is due to because in test creates first an amount and the margin is calculated wit this amount. After that, we create the last invoice and calculates the margin between the product wich margin is 500 and the amount -100, so the margin is 400 and the percent is 44.44.

account_invoice_margin/__manifest__.py
Reyes4711-S73: so we should depend on this module?
account_invoice_margin_sale depends on this module

pedrobaeza: No, this module shouldn't depend on sale on tests.

account_invoice_margin/tests/test_account_invoice_margin.py
pedrobaeza: I think this comes from not assigning the manual purchase price as it was before

account_invoice_margin/__manifest__.py
Reyes4711-S73: So we must modify the test and delete the sale.order

dreispt (approved):

dreispt:
@Reyes4711 I see warnings in the Odoo log:
```
WARNING odoo.fields: Field account.move.margin: unknown parameter 'digits', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it 
WARNING odoo.fields: Field account.move.margin_signed: unknown parameter 'digits', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it 
```

account_invoice_margin/__manifest__.py
```diff
@@ -1,10 +1,10 @@
 # Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```
pedrobaeza: Why are you changing the license? We don't want that.

account_invoice_margin/report/account_invoice_report.py
```diff
@@ -11,4 +11,4 @@ class AccountInvoiceReport(models.Model):
 
     def _select(self):
         select_str = super()._select()
-        return "%s, SUM(line.margin_signed) AS margin" % select_str
+        return "%s, line.margin_signed AS margin" % select_str
```
pedrobaeza, line 14: An operator is needed as this is a measure that can be aggregated, being `SUM`, `AVG`, or other.

account_invoice_margin/tests/test_account_invoice_margin.py
```diff
@@ -100,57 +100,3 @@ def test_invoice_refund(self):
         new_invoice = self.env["account.move"].browse(action["res_id"])
         self.assertEqual(new_invoice.invoice_line_ids.margin, 1000.00)
         self.assertEqual(new_invoice.invoice_line_ids.margin_signed, -1000.00)
-
-    def test_invoice_down_payment(self):
-        SaleAdvancePaymentInv = self.env["sale.advance.payment.inv"]
```
pedrobaeza: Why are you removing this test? It's valuable.

pedrobaeza (changes requested):

account_invoice_margin/tests/test_account_invoice_margin.py
Reyes4711-S73: This test needs  sale dependency

pedrobaeza: OK, I see. I suppose we need to include it in `account_invoice_margin_sale`.

account_invoice_margin/report/account_invoice_report.py
Reyes4711-S73: Hi @pedrobaeza,
If we add the `SUM` operator, we need to add the group_by clause because in this version, the group_by was removed.
You could see that in this PR: https://github.com/odoo/odoo/commit/af2272ac56788244a165d5b064d6c867caaf7a92#diff-ff84262b2666fce15620cbaf0c216348f92e178b752e0154ce57588e9e276bea
Could we add this in our module?
In other case, It's possible to report odoo to add again the group_by clause?
I'm waitaing you response.
Thanks for your review.

pedrobaeza: OK, I see. It seems the goal is to cache it and improve performance. But then I think we need to do something similar as in:

https://github.com/odoo/odoo/commit/af2272ac56788244a165d5b064d6c867caaf7a92#diff-2da964c74daf61be62254f9b949ed334551abfab4f9c1484d6812ccddc716f4aR149

Reyes4711-S73: Hi @pedrobaeza,
I review with two runbot (13.0 and 14.0) and both are working well. I think the sum is useless now.

pedrobaeza (approved):
OK, let's continue then

PR #92, opened 2021-04-23
[12.0][ADD] repair margin module

sergiocorato:


marcelsavegnago:
@sergiocorato  Considering that the operations are rendered services and that they can have a defined cost, I think it makes sense to include the margin fields also for the operations (repair.fee). What do you think ?

sergiocorato:
> @sergiocorato Considering that the operations are rendered services and that they can have a defined cost, I think it makes sense to include the margin fields also for the operations (repair.fee). What do you think ?

I agree, I'll extend the PR

PR #96, opened 2021-06-08
[ADD] (12.0) product_pricelist_margin - new module to preview prices and margins, depending on the pricelist

legalsylvain:
### Description

This module is designed to extend Product Module in Odoo to preview product margin
depending on the different pricelists.

### Usage

* Go to "Sale > Products > Products"

* Open a product form

* a new button is available, named "See Margins per Pricelist"

product_template_form

* Click on the button, and check if the margins are correct

wizard_preview_pricelist_margin_form

* you can optionnaly change date, or quantity, if you have complex pricelist depending on such parameters.

### Install

**Lazy Dependency**

If you install the extra module ``web_tree_dynamic_colored_field``, present in the OCA repository (https://github.com/OCA/web), the lines will have a background, depending on the final margin.

* the greener the line, the greater the margin
* the bluer the line, the lower the margin
* In case of negative margin, the line is red

with_module_web_tree_dynamic_colored_field

CC : @quentinDupont, @sandieFavre, (apps/deck/#/board/144/card/1623)


product_pricelist_margin/wizards/wizard_preview_pricelist_margin_line.py
```diff
@@ -0,0 +1,56 @@
...
+        digits=dp.get_precision('Product Price'), readonly=True,
+    )
+
+    bg_color = fields.Char(compute="_compute_bg_color")
+
+    def _compute_bg_color(self):
+        for line in self:
+            if line.margin_percent < 0:
+                line.bg_color = "#FF3333"
+            else:
```
quentinDupont, line 53: niice :ok_hand:

product_pricelist_margin/wizards/wizard_preview_pricelist_margin.py
```diff
@@ -0,0 +1,98 @@
...
+            to_currency = pricelist.currency_id
+            res = product.taxes_id.compute_all(
+                product.price,
+                to_currency,
+                self.product_qty,
+                product=product
+            )
+
+            price_vat_excl = res["total_excluded"] / self.product_qty
+            price_vat_incl = res["total_included"] / self.product_qty
```
JordiMForgeFlow, line 81: Here it should be checked that the product quantity is > 0, otherwise it triggers an error.

legalsylvain: Thanks ! 
Done. (i added ``if not self.product_id or not self.product_qty:`` at the beginning of the function.

JordiMForgeFlow: Nice, thank you! Just one question: in the margin lines displayed in the wizard, why should them be editable?

legalsylvain: I dont understand. There are not editable. Line_ids is set as readonly

JordiMForgeFlow: I mean when clicking on one line for example, the form view is opened and the user can click on Save & New, and then a blank line gets added in the form and it gets a bit weird.

image

image

legalsylvain: That's weird and seems to be a odoo bug. As the field is readonly... do you know how to prevent it ?

JordiMForgeFlow: I think that it is because there is no form view for the wizard line model, so Odoo generates this one. I think that if you define a form view for the wizard lines with just the fields that are important for the user then it should be fine. Also, I think that you should do the same you did for the main wizard form view, adding explicitly the footer with the "Close" button, otherwise the other default standard buttons will appear :)

legalsylvain: Thanks ! 
Well, I tried some ``editable="false" edit="false" create="false" delete="false"`` without success.
Your solution works ! Could you review it ?
Well, it's a pity to have to add 15 useless xml lines just to avoid a "create and new" button on a readonly field. But well !

JordiMForgeFlow (approved):
Functional and code review 👍🏼 Nice addition!

dreispt:
Travis is red

product_pricelist_margin/wizards/wizard_preview_pricelist_margin.py
```diff
@@ -0,0 +1,98 @@
...
+    @api.depends("product_id", "product_qty", "price_date")
+    def _compute_line_ids(self):
+        pricelists = self._get_pricelists()
+        lines_vals = []
+        if not self.product_id or not self.product_qty:
+            return
+        for pricelist in pricelists:
+            product = self.product_id.with_context(
+                quantity=self.product_qty,
+                pricelist=pricelist.id,
```
sergiocorato, line 69: I'm not able to have the prices changed by date, maybe the date has to be passed in context?
```suggestion
                pricelist=pricelist.id,
                date=self.price_date,
```

legalsylvain:
hi @sergiocorato Thanks for your test and your suggestion.

your patch worked, I committed it. (and added test regarding date)

could you update your review ? 

thanks !

GSLabIt:
@legalsylvain is this ready to merge?

legalsylvain:
> @legalsylvain is this ready to merge?

Yes.

PR #98, opened 2021-06-28
[14.0][ADD] sale_margin_delivery_cost: include other costs in sales margins

dreispt:


sale_margin_delivery_cost/models/sale_order.py
```diff
@@ -0,0 +1,67 @@
...
+        states={"cancel": [("readonly", True)], "done": [("readonly", True)]},
+        copy=True,
+        auto_join=True,
+    )
+
+    @api.depends("order_line.margin", "amount_untaxed", "other_cost_ids.price_unit")
+    def _compute_margin(self):
+        """
+        Subtract other costs from the Sales Order margin
+        """
```
ivantodorovich:
```suggestion
        # Override. Subtract other costs from the Sales Order margin
```

When overriding, it's better not to overwrite original method's docstrings
(there's another case just below)

ivantodorovich (approved):
LG 👍🏻

sale_margin_delivery_cost/models/sale_order.py
dreispt: AFAIU the docstrings are concatenated. I think I remember seeing that when using `help()` in the REPL.
This seems corroborated by answer #3 here https://discuss.dizzycoding.com/inheriting-methods-docstrings-in-python/
But I'm not sure how reliable that page is.

dreispt: Thank you for the review BTW

ivantodorovich: > AFAIU the docstrings are concatenated. I think I remember seeing that when using `help()` in the REPL. This seems corroborated by answer #3 here https://discuss.dizzycoding.com/inheriting-methods-docstrings-in-python/ But I'm not sure how reliable that page is.

I've just tested on `3.8.5`.. unfortunately no, docstrings are still being overwritten :( it would've been nice though
Anyways, usually commenting the reason for an override is no docstring material 😅

dreispt:
@ivantodorovich Docstrings fixed!

ivantodorovich (approved):

PR #101, opened 2021-07-26
[12.0][FIX] sale_margin_delivered: Force company for SO in tests

joao-p-marques:
The default is not being applied because of the `new`/`create` approach.
If some other module in the chain needs that field, tests will fail.
This forces the same company that is applied by default: https://github.com/odoo/odoo/blob/12.0/addons/sale/models/sale.py#L228

@Tecnativa
TT31040

ping @pedrobaeza @sergio-teruel

### DIFF
--- a/sale_margin_delivered/tests/test_sale_margin_delivered.py
+++ b/sale_margin_delivered/tests/test_sale_margin_delivered.py
@@ -39,7 +39,8 @@ class TestSaleMarginDelivered(SavepointCase):
         })
 
     def _new_sale_order(self):
-        sale_order = self.SaleOrder.new({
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
This is not the proper approach. Do something like:

```python
vals = self.env["sale.order"].default_get()
vals.update({
    ....
})
sale_order = self.SaleOrder.new(vals)
```

joao-p-marques:
Thank you for the suggestion @pedrobaeza! Changes made :+1:

pedrobaeza (approved):

pedrobaeza:
Fast-tracking it:

/ocabot merge patch

PR #102, opened 2021-08-12
[12.0][ENH] account_invoice_margin: add 'Margin(%)' field on Invoice analysis report

bodedra:
Taken screenshot from Runbot. http://3498077-102-f1859b.runbot1-2.odoo-community.org/web#action=273&model=account.invoice.report&view_type=pivot&menu_id=118

Screenshot 2021-08-13 at 11-33-15 Invoices Analysis - Odoo

PR #103, opened 2021-08-13
[13.0][ENH] account_invoice_margin: add 'Margin(%)' field on Invoice analysis report

bodedra:
Fp: https://github.com/OCA/margin-analysis/pull/102

### DIFF
--- a/account_invoice_margin/report/account_invoice_report.py
+++ b/account_invoice_margin/report/account_invoice_report.py
@@ -8,7 +8,11 @@ class AccountInvoiceReport(models.Model):
     _inherit = "account.invoice.report"
 
     margin = fields.Float(string="Margin", readonly=True)
+    margin_percent = fields.Float(string="Margin (%)", readonly=True)
 
     def _select(self):
         select_str = super()._select()
-        return "%s, SUM(line.margin_signed) AS margin" % select_str
+        return (
+            "%s, SUM(line.margin_signed) AS margin, "
+            "AVG(move.margin_percent)  AS margin_percent" % select_str
+        )
--- a/account_invoice_margin/report/account_invoice_report.py
+++ b/account_invoice_margin/report/account_invoice_report.py
@@ -36,7 +36,7 @@ class AccountInvoiceReport(models.Model):
         full_fields = all(x in fields for x in {"margin", "margin_percent"})
         margin_percent = "margin_percent" in fields
         for line in res:
-            if full_fields:  # compute difference
+            if full_fields and line["price_total"]:  # compute difference
                 line["margin_percent"] = (
                     (line["margin"] or 0.0) / line["price_total"]
                 ) * 100
--- a/account_invoice_margin/report/account_invoice_report.py
+++ b/account_invoice_margin/report/account_invoice_report.py
@@ -1,7 +1,7 @@
 # © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class AccountInvoiceReport(models.Model):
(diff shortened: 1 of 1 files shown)


legalsylvain:
something is wrong with travis.

bodedra:
> something is wrong with travis.

just fixed it.

bodedra:
cherry-pick doesn't work. Seems alias are changed in v13.0. Let me repair this PR.

account_invoice_margin/report/account_invoice_report.py
```diff
@@ -8,7 +8,11 @@ class AccountInvoiceReport(models.Model):
...
     margin = fields.Float(string="Margin", readonly=True)
+    margin_percent = fields.Float(string="Margin (%)", readonly=True)
 
     def _select(self):
         select_str = super()._select()
-        return "%s, SUM(line.margin_signed) AS margin" % select_str
+        return (
+            "%s, SUM(line.margin_signed) AS margin, "
+            "AVG(move.margin_percent)  AS margin_percen" % select_str
+        )
```
victoralmau:
```suggestion
        return (
            "%s, SUM(line.margin_signed) AS margin, "
            "AVG(move.margin_percent)  AS margin_percent" % select_str
        )
```

victoralmau (changes requested):

account_invoice_margin/report/account_invoice_report.py
bodedra: @victoralmau comment addressed! Thanks for your review.

victoralmau (approved):

bodedra:
@victoralmau still need some correction. See full details here https://github.com/OCA/margin-analysis/pull/102#pullrequestreview-729899124

account_invoice_margin/report/account_invoice_report.py
```diff
@@ -1,14 +1,45 @@
...
+            groupby,
+            offset=offset,
+            limit=limit,
+            orderby=orderby,
+            lazy=lazy,
+        )
+        if "margin" not in fields:
+            return res
+
+        full_fields = all(x in fields for x in {"margin", "margin_percent"})
```
pedrobaeza: Shouldn't this be?

```suggestion
        full_fields = all(x in fields for x in {"price_total", "margin_percent"})
```

account_invoice_margin/report/account_invoice_report.py
```diff
@@ -1,14 +1,45 @@
...
+        margin_percent = "margin_percent" in fields
+        for line in res:
+            if full_fields:  # compute difference
+                line["margin_percent"] = (
+                    (line["margin"] or 0.0) / line["price_total"]
```
pedrobaeza: Error when `line["price_total"]` = 0

bodedra: Yes. It also problem on 12. I created invoice with 0 amount.

pedrobaeza: Put then `if full_fields and line["price_total"]:` as condition

bodedra: I like your solution compare to mine
> " ( line["price_total"] or 1.0) "

account_invoice_margin/report/account_invoice_report.py
bodedra: current version of fix working well as per this example: https://github.com/OCA/margin-analysis/pull/102#discussion_r688725059 or is there any additional example to verify?

I will do the same testing once runbot is up.

pedrobaeza: The logic is that we should have here the measures that are involved in the computation, isn't it?

bodedra: Yes, that's true. IMO both approach will give us same output. I am open to change field from `margin` to `price_total`

pedrobaeza: Let's do it for this version then.

bodedra: I have added as a separate commit.

pedrobaeza:
@bodedra I'm thinking that this can be conflicting as we are accessing as `line["margin"] or 0.0` instead of `line.get("margin", 0.0)`. Why don't we put both 3 fields in the condition or switch to that way?

bodedra:
@pedrobaeza comment addressed!

pedrobaeza (approved):

PR #104, opened 2021-08-18
[13.0][IMP] account_invoice_margin: Add Cost price measure to Invoices Analysis

victoralmau:
Add Cost price measure to Invoices Analysis (unitary cost price).

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT30685

### DIFF
--- a/account_invoice_margin/report/account_invoice_report.py
+++ b/account_invoice_margin/report/account_invoice_report.py
@@ -1,4 +1,5 @@
 # © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
The name "Purchase Price" is misleading, as it's not that one, but "Cost Price".

victoralmau:
> The name "Purchase Price" is misleading, and it's not that one, but "Cost Price".

Excuse me, you mean change the field name (cost_price) or string?

pedrobaeza:
At least the label. For the field name, follow the `sale_margin` convention. I also wonder if it's the unitary cost price or the subtotal one.

victoralmau:
I have already changed the label. It's unitary cost price.

pedrobaeza:
Then it doesn't have too much sense to be sum. I also doubt the usefulness of this field. I would close this and wait for the need of this.

PR #107, opened 2021-09-06
[13.0][FIX] sale_margin_delivered: Use Form in tests to avoid crash tests

sergio-teruel:
Crash test due to this Odoo commit
OCA/OCB@8ca10a8
The wizard lines have the field uom_id related to move.product_uom and readonly=False, if you call directly to wiz.onchage_picking_id a write in stock move is executed.

cc @Tecnativa


pedrobaeza (approved):

pedrobaeza:
Merging manually as well.

PR #110, opened 2021-09-21
[13.0][IMP] account_invoice_margin: Avoid error when trying to install module with database populated with many invoices

luiseevaquer:
Currently, when the module is installed in a database populated with invoices, the compute methods are 
executed when fields are created, if the number of records is not many, the fields are set correctly, the problem 
lies when the records of these invoice lines are a considerable count (for example > 400 thousand AML records) 
does not finish running the compute methods and the odoo server closes unexpectedly, causing the module to 
fail to install.

Selección_999(603)

In order to solve this problem, it was required to create a hook to create the fields, to avoid the execution of 
the computed methods and in this way the installation of the module culminates successfully.

Selección_999(604)

### DIFF
--- a/account_invoice_margin/__init__.py
+++ b/account_invoice_margin/__init__.py
@@ -1,5 +1,6 @@
 # © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
+from .hooks import pre_init_hook
 from . import models
 from . import report
--- a/account_invoice_margin/hooks.py
+++ b/account_invoice_margin/hooks.py
@@ -1,0 +1,56 @@
+import logging
+
+from odoo.tools import column_exists, create_column
+
+_logger = logging.getLogger(__name__)
+
+COLUMNS = (
+    ("account_move", "margin"),
+    ("account_move", "margin_signed"),
+    ("account_move", "margin_percent"),
+    ("account_move_line", "margin"),
+    ("account_move_line", "margin_signed"),
+    ("account_move_line", "margin_percent"),
+    ("account_move_line", "purchase_price"),
+)
+
+
+def pre_init_hook(cr):
+    for table, column in COLUMNS:
+        if not column_exists(cr, table, column):
+            _logger.info("Create column %s in database", column)
+            create_column(cr, table, column, "numeric")
+    cr.execute(
+        """
+        WITH am AS(
+            SELECT id FROM account_move WHERE type NOT ILIKE 'in_%'
+        )
+        UPDATE account_move_line
+            SET margin = price_subtotal, margin_signed = price_subtotal,
+                margin_percent = 100
+        FROM am
+        WHERE am.id = account_move_line.move_id
+        AND price_subtotal > 0.0;
+    """
+    )
+    cr.execute(
+        """
+        WITH aml AS(
+            SELECT
+               account_move_line.move_id,
+               SUM(account_move_line.margin) AS sum_margin,
+               SUM(account_move_line.margin_signed) AS sum_margin_signed
+            FROM account_move_line
+            INNER JOIN account_move
+            ON account_move.id = account_move_line.move_id
+            GROUP BY account_move_line.move_id
+        )
+        UPDATE account_move
+            SET margin = aml.sum_margin,
+                margin_signed = aml.sum_margin_signed,
+                margin_percent = aml.sum_margin_signed / amount_untaxed * 100
+        FROM aml
+        WHERE account_move.id = aml.move_id
+        AND account_move.amount_untaxed > 0.0
+    """
+    )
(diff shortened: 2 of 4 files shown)


account_invoice_margin/__init__.py
```diff
@@ -1,5 +1,6 @@
 # © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
+from .hook import pre_init_hook
```
luisg123v: Plural in hooks.

```suggestion
from .hooks import pre_init_hook
```

account_invoice_margin/hook.py
```diff
@@ -0,0 +1,22 @@
+import logging
+
+from odoo.tools.sql import column_exists, create_column
```
luisg123v: A little bit simpler

```suggestion
from odoo.tools import column_exists, create_column
```

luisg123v (changes requested):
Looks good, however CI is red.

account_invoice_margin/hook.py
luiseevaquer: Done

account_invoice_margin/__init__.py
luiseevaquer: Done

legalsylvain (approved):
Thanks !

pedrobaeza:
I don't have it clear that this is a good solution, as it's losing the history. @sergio-teruel do you see it correct? Anyways, this is not a [FIX], but an [IMP], as your problems are due to the deployment limits.

legalsylvain:
Hi Pedro. The previous implémentation was loosing History too. In fact recovering marging History was Never implemented.
The reason is that Guess purchase_price of old account_move_line is hard. (We have to compute the standard Price of a product in a given date).

No régression here IMHO

pedrobaeza:
OK, thanks for the explanations. My colleague will confirm if there's a middle point, that is the specialist on this (maybe through the sale source...)

legalsylvain:
One point however. 

Not tested : @lescobarvx : could you take a look ?

**Before that patch / with a demo database with a sale invoice**

- purchase_price = 0
- margin = amount_untaxed
- margin_rate = 100%

**After that patch / with a demo database with a sale invoice**

- purchase_price = 0
- margin = 0
- margin_rate = 0

Am I right ? If yes, this patch introduces inconcistency. (if purchase_price is 0, margin_rate is 100%)

I think that after creating column we should execute some ```update account_move_line set margin = amount_untaxed, margin_rate = 100```. the same for ```account_move```.
maybe not the right request, I just quick written that.

sergio-teruel:
Ok. If you create the columns before the compute method is not executed, ok, all fields has zero value..
If we not create the fields, all history aml are computed with a incorrect standard_price... So this is not a good idea too ... 

Maybe would be possible do a script which get purchase_price from sale_line (if has one linked)... ???

Actually... I don't know what the best solution.

luisg123v:
@sergio-teruel from what you comment, it seems to me the history-related problem is a design one, because a computed field depends on a non-persistend field (the standard price).

If the sales price is to be used, maybe that approach should be implemented in the original compute, too.

hugho-ad (approved):

luisg123v (approved):
@moylop260 what do you think?

moylop260:
Please, check the @legalsylvain comment: 
 - https://github.com/OCA/margin-analysis/pull/110#issuecomment-924692745

@sergio-teruel  
Currently we have `killed` error installing this module from Odoo-sh

This PR is trying to allow to install this module with a similar behaviour not IMProvements here only fixing this issue

I think it can be done after fixing the comment

Are you agree?

hbto:
@lescobarvx Looks Good To Me.

Regards.

legalsylvain (changes requested):
I think you should initialize margin to 100% to avoid inconsistency. Otherwise +1 to accept this PR as it and work in a second time  regarding thé possibility to Guess purchase Price via History.

account_invoice_margin/hooks.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+        UPDATE account_move
+            SET margin = aml.sum_margin,
+                margin_signed = aml.sum_margin_signed,
+                margin_percent = aml.sum_margin_signed / amount_untaxed * 100
+        FROM aml
+        WHERE account_move.id = aml.move_id
+        AND account_move.amount_untaxed > 0.0
+    """
+    )
```
luiseevaquer, line 56: @legalsylvain @luisg123v 

What do you think?

legalsylvain: Not tested but lgtm. Thanks !

legalsylvain (approved):
Thanks !

sergio-teruel (approved):

pedrobaeza:
Please fw-port it to v14 as well, thanks.

PR #112, opened 2021-09-29
[ADD] sale_margin_mrp: New module

chienandalu:
Recompute kits and manufactured products standard price when the margin is
computed.

cc @Tecnativa TT32040

please review @sergio-teruel @carlosdauden


sergio-teruel:
@chienandalu fix travis... 
maximum recursion depth exceeded while calling a Python object :smile:

chienandalu:
It should be working now.

sergio-teruel:
:smile_cat: still fails

chienandalu:
Ouch. Silly lint error. Fixed

sergio-teruel (approved):

CarlosRoca13 (approved):

CarlosRoca13 (commented):
Ups, travis is red

sale_margin_mrp/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.onchange("product_id", "product_uom")
+    def product_id_change_margin(self):
+        """Action bom cost computes the standard price from the BoM components.
+        We reproduce here the condition that makes the manual action visible
+        or not in the product form to trigger it automaticaly."""
```
ivantodorovich, line 13:
```suggestion
        Override. Action bom cost computes the standard price from the BoM components.
        We reproduce here the condition that makes the manual action visible
        or not in the product form to trigger it automaticaly.
```

Nitpicking here: When overwritting, it's better not to overwrite the original method's docstring

sale_margin_mrp/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+        if (
+            self.product_id and self.product_id.bom_count
+            and self.product_id.valuation != "real_time"
+        ):
+            product = self.product_id.with_context(
+                force_company=self.order_id.company_id.id
+            ).sudo()
+            product.write({"standard_price": product._get_price_from_bom()})
```
ivantodorovich, line 21: Is this really what you want?

It's a bit invasive to write on the product everytime a product is added to the SO I believe.


Shouldn't this be writing on the sale.order.line's `purchase_price` instead? If so you could probably override `_compute_margin` instead, and achieve the same in a cleaner way

pedrobaeza:
Closing as not finally desired.

PR #114, opened 2021-09-29
[14.0][IMP] account_invoice_margin: Avoid error when trying to install module with database populated with many invoices

luiseevaquer:
Currently, when the module is installed in a database populated with
    invoices, the compute methods are executed when fields are created,
    if the number of records is not many, the fields are set correctly,
    the problem lies when the records of these invoice lines are a
    considerable count (for example > 400 thousand AML records) does not
    finish running the compute methods and the odoo server closes
    unexpectedly, causing the module to fail to install.

    In order to solve this problem, it was required to create a hook to
    create the fields to avoid the execution of the computed methods
    and initialize margin percent to 100% to avoid inconsistency, so
    that in this way the installation of the module culminates successfully.

### DIFF
--- a/account_invoice_margin/__manifest__.py
+++ b/account_invoice_margin/__manifest__.py
@@ -18,4 +18,5 @@
         "security/account_invoice_margin_security.xml",
         "views/account_invoice_margin_view.xml",
     ],
+    "pre_init_hook": "pre_init_hook",
 }
--- a/account_invoice_margin/__init__.py
+++ b/account_invoice_margin/__init__.py
@@ -1,5 +1,6 @@
 # © 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from .hooks import pre_init_hook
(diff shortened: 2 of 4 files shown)


account_invoice_margin/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Account Invoice Margin",
     "summary": "Show margin in invoices",
-    "version": "14.0.1.0.0",
+    "version": "14.0.1.0.1",
```
legalsylvain: for your information, such change is useless 
when merging with @OCA-git-bot, the bump of the version will be done. (if asked by maintainer).

luiseevaquer: I got it, thanks

moylop260 (approved):
👍

luisg123v (approved):
@pedrobaeza this is the forward-port of https://github.com/OCA/margin-analysis/pull/110 you asked for.

CC @hugho-ad

pedrobaeza (approved):
Thanks

PR #120, opened 2021-10-28
[12.0][FIX] sale_margin_delivered: Margin delivered is not computed when price_reduce is zero (Discount of 100%)

sergio-teruel:
cc @Tecnativa

ping @carlosdauden @chienandalu

### DIFF
--- a/sale_margin_delivered/models/sale_margin.py
+++ b/sale_margin_delivered/models/sale_margin.py
@@ -30,7 +30,7 @@ class SaleOrderLine(models.Model):
                  'move_ids.price_unit')
     def _compute_margin_delivered(self):
         digits = self.env['decimal.precision'].precision_get('Product Price')
-        for line in self.filtered('price_reduce'):
+        for line in self:
             if not line.qty_delivered and not line.product_uom_qty:
                 continue
             line.margin_delivered = 0.0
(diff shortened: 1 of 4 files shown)


sale_margin_delivered/models/sale_margin.py
```diff
@@ -55,6 +55,8 @@ def _compute_margin_delivered(self):
                 )
             # compute percent margin based on delivered quantities or ordered
             # quantities
-            line.margin_delivered_percent = line.qty_delivered and (
+            line.margin_delivered_percent = (
+                line.qty_delivered and line.price_reduce and (
                 (line.price_reduce - line.purchase_price_delivery) /
                 line.price_reduce * 100.0) or 0.0
+            )
```
chienandalu, line 62: Shouldn't  the percentage be negative? Something like:

cost: 5 discount: 100% margin: -5 percentage: -100%

sergio-teruel: We can set to 0% because you are selling with zero price... So the margin can not be computed... so is 0.0%

carlosdauden (approved):
Any division by 0 results in infinity.
In this specific case, you could put -100% for a matter of taste, but generally in Odoo it usually shows 0.
This data is not significant when the sale amount is 0, so I prefer to show 0, since -100 is a valid value that can be obtained when we sell something for 2 at a cost of 4.

carlosdauden:
Please, fix unrelated issue:
https://app.travis-ci.com/github/OCA/margin-analysis/jobs/545554564#L321

sergio-teruel:
Done in separated commit

chienandalu (approved):
ok then :+1:

pedrobaeza:
Please fw-port it to v13:

/ocabot merge patch

pedrobaeza:
@sergio-teruel was this included in 13.0?

chienandalu:
The merge didn't success

pedrobaeza:
Yeah, but this branch is broken, and we are not using it anymore. I just merged for having the patch, but what we are now interested in is to have the patch in upper versions.

chienandalu:
Ok, I'll work on the forwards

PR #122, opened 2021-12-15
[14.0][MIG] product standard margin

cvinh:
nothing to do


product_standard_margin/readme/ROADMAP.rst
```diff
@@ -0,0 +1,7 @@
+* This module will not work properly if used in a multicompany context with product
+  prices depending on the company.
+
+* It should be great to use the odoo widget ``percentpie`` for the display of the
+  field ``standard_margin_rate`` but for the time being, the field is not displayed
+  properly.
+  Ref of the Odoo Issue: https://github.com/odoo/odoo/issues/42869
```
legalsylvain: Did you tested if this bug was solved ?

legalsylvain:
Thnaks for porting this module.
Could you check pre-commit ?

product_standard_margin/readme/ROADMAP.rst
cvinh: done it works in v14

cvinh:
Also I added markup rate

cvinh:
@legalsylvain any chance to merge ?

product_standard_margin/readme/ROADMAP.rst
legalsylvain: Hi. Just tested, it doesn't seems to work. Or did I missed something ?

legalsylvain: image

legalsylvain: Well, the bug in 14.0 still remains. However it's a pity to not use this great widget. 

Analysing : 
A text is displayed after the percent graph, because of : 
https://github.com/odoo/odoo/blob/14.0/addons/web/static/src/xml/base.xml#L1370

This line is defined because they are used in a widget that should be defined as ``statinfo``, but both are incompatible.

https://github.com/odoo/odoo/blob/14.0/addons/mass_mailing/views/mailing_mailing_views.xml#L125

Not sure that it will be fixed in odo,o because this widget is used only in one module, in one form. (``mass_mailing``)

**Option 1**
- conserve the use of ``percentpie`` widget
- keep an updated text to say that there is a bug to be fixed in odoo. (that can be fixed, removing a line in the ``base.xml`` file).

**Option 2**
- remove the use of ``percentpie`` widget.

What do you think ?

cvinh: Option 1 : we have customers who like the percentpie ;-)
BTW we even did not notice that the extra text was a bug... it's super minor to us

legalsylvain: >  BTW we even did not notice that the extra text was a bug... it's super minor to us

;-)

ok for me, not perfect solution. Could you just re add some text in the ROADMAP.rst section to mention that minor display bug ?

legalsylvain (approved):
One minor remark regarding rst file. Otherwise, LGTM. 
code review and test.

thanks for porting this module ! 

Note : I have no write access on that repo. (even if I'm marked as the maintainer of this module).

cvinh:
I did a rebase but it did not change that pr. I'm doing another one

PR #124, opened 2022-01-03
[FIX] sale_margin_delivered: non storable

chienandalu:
We don't want to compute margin based on stock moves for non storable
products. This specially true for services and kits.

cc @Tecnativa TT33499

What do you think @sergio-teruel @carlosdauden ?

### DIFF
--- a/sale_margin_delivered/__manifest__.py
+++ b/sale_margin_delivered/__manifest__.py
@@ -9,7 +9,7 @@
     "category": "Sales",
     "license": "AGPL-3",
     "depends": [
-        "stock",
+        "sale_stock",
         "sale_margin_security",
     ],
     "data": [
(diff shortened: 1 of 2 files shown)


pedrobaeza:
In the case of services, there's also a delivered quantity to be considered, for being margins when you work on fixed rate, but the time spent is variable.

chienandalu:
Yes, and that'd be covered by this change

pedrobaeza:
Not sure why Travis fails.

sergio-teruel:
Test fails due to product_standard_margin module...

sergio-teruel:
@chienandalu Can you add any test to cover this... ??

sergio-teruel:
@chienandalu Sorry... Can you change dependendy from sale to sale_stock??

chienandalu:
It's already depending on sale and stock: https://github.com/OCA/margin-analysis/blob/367c4aff6c098b55dfe7cdc8e86b079815e19608/sale_margin_delivered/__manifest__.py#L12-L13

sergio-teruel (approved):

pedrobaeza (approved):

carlosdauden (approved):

PR #125, opened 2022-02-04
[14.0][MIG] sale_order_percent: Migration to 14.0

Reyes4711-S73:
Standard migration to 14.0


legalsylvain (changes requested):
This module is obsolete in v14. (See roadmap)
https://github.com/odoo/odoo/blob/14.0/addons/sale_margin/models/sale_order.py#L12

However openupgrade script should be done (if not yet) to rename margin_percent into percent.

Kind regards.

pedrobaeza:
Scripts should be done in OU itself, detecting if the column exists previously in pre of `sale_margin` module

PR #130, opened 2022-05-19
[14.0] [IMP] `account_invoice_margin`: several small improvements

ivantodorovich:
Easier to review commit by commit

### DIFF
--- a/account_invoice_margin/report/account_invoice_report.py
+++ b/account_invoice_margin/report/account_invoice_report.py
@@ -7,8 +7,22 @@ from odoo import fields, models
 class AccountInvoiceReport(models.Model):
     _inherit = "account.invoice.report"
 
-    margin = fields.Float(string="Margin", readonly=True)
+    margin = fields.Float(readonly=True)
+    purchase_price = fields.Float("Cost", readonly=True)
 
     def _select(self):
-        select_str = super()._select()
-        return "%s, line.margin_signed AS margin" % select_str
+        res = super()._select()
+        outbound_types = tuple(self.env["account.move"].get_outbound_types())
+        return ",\n".join(
+            [
+                res,
+                "line.margin_signed AS margin",
+                f"""
+                (
+                    line.purchase_price
+                    * line.quantity
+                    * (CASE WHEN move.move_type in {outbound_types} THEN -1 ELSE 1 END)
+                ) AS purchase_price
+                """,
+            ]
+        )
--- a/account_invoice_margin/models/account_invoice.py
+++ b/account_invoice_margin/models/account_invoice.py
@@ -85,7 +85,7 @@ class AccountMoveLine(models.Model):
         self.ensure_one()
         return self.product_id.standard_price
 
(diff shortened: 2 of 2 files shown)


account_invoice_margin/report/account_invoice_report.py
```diff
@@ -7,8 +7,22 @@
 class AccountInvoiceReport(models.Model):
     _inherit = "account.invoice.report"
 
-    margin = fields.Float(string="Margin", readonly=True)
+    margin = fields.Float(readonly=True)
+    purchase_price = fields.Float("Cost", readonly=True)
```
legalsylvain, line 11: naive question (not blocking) : ``margin`` is ``Monetary`` on ``account.move.(line)`` and ``Float`` on report. is it attended ?

legalsylvain (approved):
Thanks a lot for all that changes ! 

LGTM. Code review / no test.

account_invoice_margin/report/account_invoice_report.py
ivantodorovich: I tried to use Monetary here, but then odoo started complaining about not knowing the currency field 
As I couldn't find any Monetary fields on this report, I left it as Float 😅 I didn't dig much tbh

ivantodorovich:
Just rebased

account_invoice_margin/models/account_move_line.py
```diff
@@ -4,55 +4,6 @@
 from odoo import api, fields, models
 
 
-class AccountMove(models.Model):
```
pedrobaeza: This change is not correct, as it was that way on purpose, as models coming from one2many are allowed and even recommended to be kept on the same file.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#file-naming

account_invoice_margin/models/account_move_line.py
```diff
@@ -48,7 +48,7 @@ def _get_purchase_price(self):
         self.ensure_one()
         return self.product_id.standard_price
 
-    @api.depends("product_id", "product_uom_id")
+    @api.depends("product_id", "product_uom_id", "move_id.move_type")
```
pedrobaeza: Is it possible to change the type of an invoice "on the fly"?

account_invoice_margin/models/account_move_line.py
ivantodorovich: Oh, I always forget about that. My personal taste differs here.
Anyway, reverted.

account_invoice_margin/models/account_move_line.py
ivantodorovich: In odoo standard: action_switch_invoice_into_refund_credit_note
In any case, better to have the field dependencies. Even if you don't expect them to change, some other module might need it ^^

sergio-teruel (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to v15 if needed.

PR #135, opened 2022-07-06
[12.0][FIX] product_pricelist_margin : change dependency 

legalsylvain:
following @pedrobaeza : I checked and installed the module. ``product_pricelist_margin``. changing the dependency from 'sale' to 'account' is working. (https://github.com/OCA/product-attribute/pull/1101#issuecomment-1176251703)

### DIFF
--- a/product_pricelist_margin/__manifest__.py
+++ b/product_pricelist_margin/__manifest__.py
@@ -10,7 +10,7 @@
     "maintainers": ["legalsylvain"],
     "website": "https://github.com/OCA/margin-analysis",
     "license": "AGPL-3",
-    "depends": ["sale"],
+    "depends": ["account"],
     "data": [
         "views/view_product_template.xml",
         "wizards/wizard_preview_pricelist_margin.xml",


pedrobaeza:
But I want a pure `product` dependency, which is possible for only getting the price. Why do you need `account`?

legalsylvain:
Account is required for computing the Margin. (Compute-all function)

I'm curious, in witch case do you use pricelist without account installed ?

pedrobaeza:
It's not a real case right now, but to reduce the dependencies to minimum, the same as Odoo does with the pricelists being bundled in `product` module, not `sale` nor `account`. What about putting in 13.0 our module as is (we keep your commit history and attribution though) and mount on top of it this extension for margin?

pedrobaeza (approved):

legalsylvain:
> What about putting in 13.0 our module as is (we keep your commit history)

I don't understand how it can be "your module", but "my history". But maybe I missed something...

> It's not a real case right now, but to reduce the dependencies to minimum

Well, I think it will be allways a theoritical case. (using pricelist without account, because pricelist is used in website_sale, sale and point_of_sale, that all depends (of course) of account), so arguing theoritical case that can not occures is maybe not a valid argument ?). 

We have two options : 
1) one single module with margin and all the stuff. (= port current OCA ``product_pricelist_margin`` in V13, adapting with your remarks)  
2) two modules, one with price feature, and another glue module with margin.

I think having 2 modules : 
- is more work to maintain
- make margin feature (that are interesting for end user) less visible.

So, that giving said, I don't make it a casus belli.

Let me know.

pedrobaeza:
I take your commit history, rewriting the module name, and then just add a new commit with the changes to make it work according new scheme for 13.0.

legalsylvain:
OK. (Maybe you didn't saw my previous edited comment)

But If I reword your proposal : 

- you port the module ``product_pricelist_margin`` from 12.0 to 13.0, cherry picking my commits
- you rename it into ``product_pricelist_simulation``
- you move the module into ``OCA/product-attribute``
- you remove all the margin stuff

- I add a ``ROADMAP.rst`` section in ``product_pricelist_margin`` in the V12 release
- when I port the module, I add a dependency to ``product_pricelist_simulation`` and remove all the stuff that is present in ``product_pricelist_simulation``
- I rename the module into ``product_pricelist_simulation_margin``
- I add a rename line in the apriori.py of the Openupgrade V13 project to say that ``product_pricelist_margin`` has been renamed into ``product_pricelist_simulation_margin``.

even if it requires a bit more work than keeping a single module, I can live with it, and if it allows the OCA community to make more use of the simulation module developed by GRAP, that's great in the long run.

Let me know if all is OK for you.

pedrobaeza:
Great, thanks for arriving to a deal. We'll do in the following days.

PR #137, opened 2022-07-28
[15.0][MIG] account_invoice_margin_sale: Migration to 15.0

victoralmau:
Migration to 15.0

Please @sergio-teruel and @carlosdauden could you review it?

@Tecnativa TT36420


account_invoice_margin_sale/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,8 @@
+* `Tecnativa <https://www.tecnativa.com>`__:
+
+  * Sergio Teruel
+  * Carlos Dauden
```
sergio-teruel, line 4: @victoralmau, add yourself as a contributor

sergio-teruel (approved):
Minor comment

account_invoice_margin_sale/models/account_invoice.py
```diff
@@ -15,6 +15,7 @@ def _get_margin_applicable_lines(self):
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
+    # pylint: disable=W8110
```
pedrobaeza, line 18: This shouldn't be needed, as you are calling super.

victoralmau: This is necessary because currently the super is only called on some lines.

Another possibility is to call the super at the beginning and then update the margin of the lines as needed.

What do you think about this?

pedrobaeza: Yeah, this is on purpose, but the linter fails with this as is?

victoralmau: This comment is now necessary to ensure that the linter does not fail.

pedrobaeza (approved):

PR #139, opened 2022-08-15
[14.0][FIX] account_invoice_margin: Take currency into account

pedrobaeza:
The invoice may be done in a currency different from the company currency, so the cost (purchase_price) amount will be incorrect in such context.

Forward-port of #138 

@Tecnativa TT38500


pedrobaeza:
Fast-tracking as already merged in previous version:

/ocabot merge patch

PR #169, opened 2023-04-27
[16.0][MIG] product_margin_classification

legalsylvain:


legalsylvain:
@vdewulf : could you take a look on that module ? I remember that you had another implementation of that feature. convergence could be great !

vdewulf:
@victor-champonnois Can you follow up this topic please with Sylvain?
I'll give you the context of our discussion in April in Lyon.
Thanks!

product_margin_classification/data/decimal_precision.xml
```diff
@@ -0,0 +1,14 @@
...
+Copyright (C) 2016-Today GRAP (http://www.grap.coop)
+@author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+<odoo noupdate="1">
+
+    <record forcecreate="True" id="decimal_margin" model="decimal.precision">
+        <field name="name">Margin Rate</field>
+        <field name="digits">4</field>
+    </record>
```
amarcosg, line 12: Why do you create your own decimal precision here in this module? I ask because in another modules of this repo we are still using "Product Price". Wouldn't it make sense to have the same precision for all margin fields?

legalsylvain: Hi thanks for the review. Well. Product price precision and Margin precision have nothing to do.
product price is used for the field expressed in currency (like the base field ``lst_price`` ). 
Margin rate is used for percentage field.
Use case : In Europe you're selling a product with a very narrow margin (such as oil or medicines).
- Product Price decimal in 2
- Margin rate is 4 (or 6, or whatever).

I don't know exactly which module you are mentionning. account_invoice_margin margin_percent field ? 
(https://github.com/OCA/margin-analysis/blob/15.0/account_invoice_margin/models/account_invoice.py#L24)
Here is it is a "displayed margin". (computed margin). So if it is rounded, it's maybe not a big deal. But in the current case, this field is used to compute margin, so maybe 0.005% of margin can change the price.

2 solutions i see : 
- we keep that PR as it.
- I create a trivial module ``product_margin_rate`` that creates that record, and I make ``product_margin_classification``  depends on  ``product_margin_rate``. (and maybe other OCA modules).

Let me know !

amarcosg: Hi @legalsylvain! Thanks for the explanation, the field makes perfect sense as you explained and made it so much clear.

The modules i was talking about are the 2 which i opened a PR to 16.0 yesterday:

- product_replenishment_cost
- product_standard_margin

They both use precision="Price Unit" in some fields (although it would only apply to some fields in product_standard_margin).  They are only for display purpose, so we are okey having the current digits so the new module is not needed.

Thank you!

victor-champonnois (approved):
Functionnal review OK.

- Some comments : why isn't the field "Theorical Margin (%)" called Markup rate (as it is called in the field description) ? That would make more sense I think.  

- I tried to reproduce the computation of margin based on markup (and the other way around), but I failed (which was expected, I suck at math). Could you add in the code, or in the readme an explanation of the math, or a relevant link ?
`````
classification.markup = 100 * (
                classification.profit_margin / (100 - classification.profit_margin)
 `

legalsylvain:
Hi @victor-champonnois.

> Some comments : why isn't the field "Theorical Margin (%)" called Markup rate (as it is called in the field description) ? That would make more sense I think.

Sorry, I think there is a misunderstood here. this module ``product_margin_classification`` doesn't introduce the field ``Theorical Margin (%)``. It is introduced by another module ``product_standard_margin``.  (https://github.com/OCA/margin-analysis/tree/15.0/product_standard_margin#product-margin-and-margin-rate) 

I'll do a PR, because indeed, the field is bad named I think.

> Could you add in the code, or in the readme an explanation of the math, or a relevant link ?

The readme contains that text : 

    This module use both [Markup](https://en.wikipedia.org/wiki/Markup_(business))
    and [Profit Margin](https://en.wikipedia.org/wiki/Profit_margin) concepts.

Here are the current help on the field : 

    markup = fields.Float(
        help="Value that help you to compute the sale price, based on your"
        " cost, as defined: Sale Price = (Cost * (100 + Markup)) / 100\n"
        "It is computed with the following formula"
        " Markup = 100 * (Sale Price - Cost) / Cost",
    )

    profit_margin = fields.Float(
        help="Also called 'Net margin' or 'Net Profit Ratio'.\n"
        "It is computed with the following formula"
        " Profit Margin = 100 * (Sale Price - Cost) / Sale Price",
    )

What do you propose ?

victor-champonnois:
@legalsylvain Ah I see that in the wikipedia links there is a formula for conversion from markup to margin that correspond to your computation, so you can ignore my comment.

dreispt (approved):

PR #171, opened 2023-05-24
[14.0][IMP] account_invoice_margin: add margin_percent to Invoice Analysis

sbiosca-s73:
Added the field margin_percent from account_move_line in the invoice analysis


dreispt (approved):

rafaelbn:
Thank you @sbiosca-s73 ! Could you please make a PR to v15 and 16? If not we are missing this in future versions 😢

PR #174, opened 2023-06-05
[16.0][MIG] product_replenishment_cost: Migration to 16.0 (from 12.0)

amarcosg:


legalsylvain (commented):
Hi. Diff review ok. Thanks for porting this module.
however, I'm asking about the usefulness of this module. Why do you need this module ? is it related to the module ``product_cost_incl_bom`` ? or other use case(s) ?

amarcosg:
Hi @legalsylvain first thanks for the review.

Ill give you and insight of the use we give it comparing it with the cost price. The cost price is the cost assigned for each product in the warehouse, the replenishment cost is the cost it would cost to buy a new product. They are different costs because the cost price depends on your valuation method and the operations you have performed while the replacement cost is determined by the current market conditions.

The price of the product in the supplier's catalog is €15/piece. Therefore, if I want to buy a new unit, my replenishment cost would be €15. But if my stock comes from having bought it in a special offer that allowed me to buy it for €10, then my cost price is €10/piece.

We use it mainly by not taking into account the cost price but calculating the replenishment cost from the suppliers' rates. 

We are going to implement this logic in a module for v16, I have no problem in proposing it for this repository if as a community we see that it is useful.

legalsylvain:
Hi  @amarcosg. 
First, thanks for your explanation. 
1) I think that your comment could be a part of the DESCRIPTION.rst file, to explain better the objective of this module. (the use case with prices, for exemple).
2) side question (not related to the current PR) : how do you update the standard_price, in your process ? I mean, if you buy 5 products at 8€ and 5 products to 12€, how do you update the standard price, to put 10€ ?

amarcosg:
@legalsylvain Ill update the DESCRIPTION.rst to made it more clear.

Regarding the second point, it depends on the customer. Some use the average price, others as in the example I have given you would correspond to the last purchase price, which is another option we have.

In that case the cost price would correspond to the last price at which I have purchased but the replenishment cost would depend on the suppliers rates of the product.

legalsylvain:
Thanks for the explanation !

rafaelbn:
Thank you!

Which is the differences with product supplierinfo?

amarcosg:
Hi @rafaelbn thank you.

In the example i give here https://github.com/OCA/margin-analysis/pull/174#issuecomment-1576827581 i only take into account the calculation with the supplierinfo, but it doesnt have to be like that always. It may be equal to our best supplierinfo in some cases (we mostly do it that way), but it may not. For example with bill of materials or some products we invoice at product reception an we use that price.

It allows us some flexibility and above all it gives us insight into the costs of our inventory.

hugo-cordoba (approved):
LGTM!

hugo-cordoba:
Hi @legalsylvain, I saw that you commented on the PR, do you see the final result well, can you approve it and merge it? 

thanks in advance, best regards!

hugo-cordoba:
Hi, sorry for the insistence @legalsylvain, can you take a look at it?

legalsylvain (approved):
thanks !

hugo-cordoba:
Hi @pedrobaeza!

Could you take a look this PR and merge please? 

Thankss! Best regards!

PR #175, opened 2023-06-05
[16.0][MIG] sale_report_margin: Migration to 16.0

flalexg:


polqubiq:
LGTM

polqubiq (approved):

JanQubiQ:
LGTM @flalexg . It seems the README is missing to be migrated. Could you check it, please?

amarcosg:
I made a branch from yours to increment coverage. if you find it right we can merge/cherry-pick the commit so it pases tests. https://github.com/factorlibre/margin-analysis/pull/4

flalexg:
> LGTM @flalexg . It seems the README is missing to be migrated. Could you check it, please?

Hi! I checked the README and I think its okey. When its merged the new README will be generated, but it has de the fragments.

amarcosg (approved):
LGTM

sale_report_margin/reports/sale_report.py
```diff
@@ -9,19 +9,13 @@ class SaleReport(models.Model):
...
-        )
+    def _select_additional_fields(self):
+        res = super()._select_additional_fields()
+        res["purchase_price"] = "l.purchase_price"
+        return res
+
+    def _group_by_sale(self):
+        res = super()._group_by_sale()
+        res += """,
+            l.purchase_price"""
```
pedrobaeza: Being this one a measure, is it needed to be grouped by it?

pedrobaeza (commented):
Please name the commit adding the tests also with the module name according guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

sale_report_margin/reports/sale_report.py
flalexg: Hi @pedrobaeza! Sorry for the late response. You are right, its unnecessary. Deleted the function. Thank you!

pedrobaeza: I still see it in the PR.

HugoCordobaLeal:
@pedrobaeza sorry, can you take a look now??

flalexg:
@pedrobaeza now I remember why I used de group by. 

psycopg2.errors.GroupingError: column "l.purchase_price" must appear in the GROUP BY clause or be used in an aggregate function
LINE 64:             l.purchase_price AS purchase_price

Shall we keep the first commit?

sale_report_margin/reports/sale_report.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class SaleReport(models.Model):
+    _inherit = "sale.report"
+
+    purchase_price = fields.Float(readonly=True)
+
+    def _select_additional_fields(self):
+        res = super()._select_additional_fields()
+        res["purchase_price"] = "l.purchase_price"
```
pedrobaeza: You have to overwrite `_select_sale` instead and do `SUM(l.purchase_price)` for having this correctly and not grouping by.

pedrobaeza: Well, maybe this can serve:

```suggestion
        res["purchase_price"] = "SUM(l.purchase_price)"
```

hugo-cordoba:
Hi @pedrobaeza!

@flalexg and I have been reviewing it, and, with the change you propose it works as expected.

Thank you!

pedrobaeza:
Great. Note that it's not the same that the module installs without error than it works as expected, and with the group by, you would find ugly totals or unfolded results when using the module.

/ocabot merge nobump

PR #178, opened 2023-07-24
[FIX] product_standard_margin

hugo-cordoba:
Hi!

After testing the module, the 'percentpie' widget does not work as expected when it has 3 digits. Moreover, in v16 it is hardly used, so I do not see the need to add it and follow the line that is followed in the version.

Captura de pantalla de 2023-07-24 11-31-50

This PR only removes the 'percentpie' widget from the views of this module.

### DIFF
--- a/product_standard_margin/views/view_product_product.xml
+++ b/product_standard_margin/views/view_product_product.xml
@@ -13,8 +13,8 @@
                 position="after"
             >
                 <field name="standard_margin" widget="monetary" />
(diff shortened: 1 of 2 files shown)


amarcosg (approved):

legalsylvain (approved):
Thanks for reporting.

/ocabot merge patch

hugo-cordoba:
Thanks @legalsylvain

Issue #180, opened 2023-08-17
Bug on product_margin_classification when variants product not activated on DB


vdewulf:
## Module
product_margin_classification

## Describe the bug

When selecting a product margin classification on a product form, the field is empty when saving manually (and if we don't save manually as not required on v16, the field is empty when moving to another screen).

## To Reproduce
**Affected versions**: v16

1. Install module, create a margin classification in the configuration
2. Go on a product form (product.template) and select a value on the product margin classification field
3. Save => the field is empty (no error is displayed)

**Expected behavior**
3. Save => the field keeps the value selected by user

**Additional context**
We investigated the issue with a developer in our team @huguesdk. Issue comes from the fact that the value of this field on the product.template is overwritten by the value on the product.product.
We think that in case there is no variants products on the database (hence 1 product.product for 1 product.template), the field should be exactly the same on both models (like the field pricelist for example).

We might propose a PR to fix this in the following months depending on the need of this module for our customers in v16.
@legalsylvain fyi

huguesdk:
@legalsylvain has this been correctly fixed by #181?

vdewulf:
I tested on runboat on v16.

Steps to reproduce the behavior:

    Install module, déactivate variants, create a margin classification in the configuration
    Go on a product form (product.template) and select a value on the product margin classification field
    Save => the field keeps the value selected by user
OK

vdewulf:
fixed

PR #181, opened 2023-08-18
[FIX] product_margin_classification : do not loose margin classification when creating new product.template. 

legalsylvain:
For that purpose, use same algorithm as 'weight' field (https://github.com/odoo/odoo/blob/16.0/addons/product/models/product_template.py#L88)

- add _inverse_margin_classification_id function
- overload _get_related_fields_variant_template

Add new test.

Fix : #180

CC : @huguesdk, @vdewulf

### DIFF
--- a/product_margin_classification/models/product_template.py
+++ b/product_margin_classification/models/product_template.py
@@ -14,7 +14,7 @@ class ProductTemplate(models.Model):
     margin_classification_id = fields.Many2one(
         string="Margin Classification",
         compute="_compute_theoretical_multi_template",
-        readonly=False,
+        inverse="_inverse_margin_classification_id",
         comodel_name="product.margin.classification",
     )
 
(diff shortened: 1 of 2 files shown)


product_margin_classification/models/product_template.py
```diff
@@ -66,6 +78,19 @@ def _compute_theoretical_multi_template(self):
...
+        if variant_count == 1:
+            self.product_variant_ids.margin_classification_id = (
+                self.margin_classification_id
+            )
+        elif variant_count == 0:
+            archived_variants = self.with_context(active_test=False).product_variant_ids
+            if len(archived_variants) == 1:
+                archived_variants.margin_classification_id = (
+                    self.margin_classification_id
+                )
```
huguesdk: i’m not sure i understand the logic here. what are the reasons to treat archived products separately? why not simply assign the margin classification to all products, archived or not?

huguesdk (commented):
thanks for the very quick response, @legalsylvain! it seems to work for us. just one comment.

product_margin_classification/models/product_template.py
legalsylvain: You're right. I first applied the same algorithm as for the _set_barcode function.
I finally used the algorithm of _set_weight. Should be OK now.

thanks for your review.

huguesdk (approved):
thanks! lgtm!

remytms (approved):
LGTM.

legalsylvain (commented):
/ocabot merge patch

PR #176, opened 2023-06-05
[16.0][MIG] sale_margin_delivered

aliciagaarzo:
Migration of **sale_margin_delivered** module to version 16

PR #183, opened 2023-09-12
[16.0][MIG] sale_margin_delivered: Migration to 16.0

pilarvargas-tecnativa:
Supersede: https://github.com/OCA/margin-analysis/pull/176

cc @Tecnativa TT44910

@chienandalu @sergio-teruel please review


sale_margin_delivered/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Sale Margin Delivered",
+    "version": "16.0.1.0.0",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/margin-analysis",
+    "category": "Sales",
+    "license": "AGPL-3",
+    "depends": ["sale_stock", "sale_margin_security"],
```
legalsylvain:
```suggestion
    "depends": ["sale_stock"],
```

Please do not force to install module that are not required.

legalsylvain (approved):
code review.
Apart a dependency question. LGTM.
thanks !

sale_margin_delivered/__manifest__.py
pilarvargas-tecnativa: But you need to depend on this module to be able to apply the permissions of the "sale_margin_security.group_sale_margin_security" group in the "sale_margin_delivered.view_order_form" view. Isn't it?

legalsylvain: can be a glue module. something like ``sale_margin_delivered_security``
- module A : sale_margin_security deploy a restriction to hide margin to some users
- module B : sale_margin_delivered displays real delivered margin. (other feature).

user can want feature B without feature A. 
Making both modules working together can be done with a glue module IMO.

carlosdauden: I agree with what you say Sylvain, someone who has the need to have those modules separately should feel free to make those changes.

sergio-teruel (approved):
LGTM!! 
Functional review ok

carlosdauden (approved):

sale_margin_delivered/__manifest__.py
legalsylvain: ok ! So I'll propose a PR against this PR.

legalsylvain (changes requested):
waiting for some refactoring in the the dependency.

sale_margin_delivered/__manifest__.py
legalsylvain: Done here : https://github.com/Tecnativa/margin-analysis/pull/2

sale_margin_delivered/views/sale_margin_delivered_view.xml
```diff
@@ -0,0 +1,24 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='margin']" position="after">
+                <field
+                    name="purchase_price_delivery"
+                    string="Cost Price dlvd."
+                    optional="hide"
+                />
+                <field
+                    name="margin_delivered_percent"
+                    string="% Margin"
```
legalsylvain:
```suggestion
                    string="% Margin dlvd. (%)"
```



This string is error prone IMO. The standard margin (introduced in sale_margin) is name "Margin (%)". 

if both fields are displayed, it is not possible for end user to know what are the field displayed. In my opinion, should "Margin dlvd. (%)"
image

legalsylvain (commented):
Functional review on runboat. One comment inline.

other question : why this module doesn't introduced delivered margin at sale.order level, next to the standard one ? 

image

pedrobaeza:
Please do a rebase to remove the merge commit.

sale_margin_delivered/views/sale_margin_delivered_view.xml
pilarvargas-tecnativa: You are right much clearer and avoids leading to errors, thanks for everything :)

sale_margin_delivered/migrations/16.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if env["ir.module.module"].search(
+        [("name", "=", "sale_margin_delivered"), ("state", "=", "installed")]
```
pedrobaeza: This is not needed, as if this is being executed is because the module is installed.

sale_margin_delivered/migrations/16.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
...
+    ):
+        sale_margin_delivered_security = env["ir.module.module"].search(
+            [
+                ("name", "=", "sale_margin_delivered_security"),
+                ("state", "=", "uninstalled"),
+            ]
+        )
+        if sale_margin_delivered_security:
+            sale_margin_delivered_security.button_immediate_install()
```
pedrobaeza: Don't do this, but just change the state to `to install`.

pilarvargas-tecnativa: corrected :)

sale_margin_delivered/migrations/16.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    sale_margin_delivered_security = env["ir.module.module"].search(
+        [
+            ("name", "=", "sale_margin_delivered_security"),
+            ("state", "=", "uninstalled"),
+        ]
+    )
+    if sale_margin_delivered_security:
+        sale_margin_delivered_security.state = "installed"
```
pedrobaeza:
```suggestion
        sale_margin_delivered_security.state = "to install"
```

legalsylvain:
Hi. I had a unanswered question here : https://github.com/OCA/margin-analysis/pull/183#pullrequestreview-1626771131
if anybody can answer... 
Thanks !

pedrobaeza:
> Hi. I had a unanswered question here : [#183 (review)](https://github.com/OCA/margin-analysis/pull/183#pullrequestreview-1626771131) if anybody can answer... Thanks !

It can be an improvement, yes

PR #192, opened 2023-12-27
[16.0][FIX]sale_margin_security: allow sale.order duplication

Tisho99:
Users without the "sale_margin_security.group_sale_margin_security" permission group can not duplicate sale.order records due to a permission error. This PR avoids that error setting Copy=False in that inherited field's inheritance.

image

Extended explanation:

The purchase_price field has:
-  A default value of Copy=True (because it is defined on the Odoo sale_margin module with readonly=False). When a user 
- A groups parameter  = 'sale_margin_security.group_sale_margin_security'

When a user without the 'sale_margin_security.group_sale_margin_security' tries to copy a record of the model, an error is shown because Odoo tries to access to the field with the salesman user but it does not have the permission

The purchase_price has a computed default value, and the case of manually editing it is not common; so setting copy=False on the field and priorizing the compute over the manual changes should not be a problem

### DIFF
--- a/sale_margin_security/models/sale_order.py
+++ b/sale_margin_security/models/sale_order.py
@@ -21,5 +21,5 @@ class SaleOrderLine(models.Model):
         groups="sale_margin_security.group_sale_margin_security"
     )
     purchase_price = fields.Float(
-        groups="sale_margin_security.group_sale_margin_security"
+        groups="sale_margin_security.group_sale_margin_security", copy=False
     )


Tisho99:
@ValentinVinagre @HaraldPanten @pedrobaeza

HaraldPanten (approved):
It's functionally working fine. Waiting for technical reviews in order to decide that this is the best way for fixing this issue.

pedrobaeza:
Although it's correct, the proper place to do this is in Odoo. In fact, we have done it already for 15.0: https://github.com/odoo/odoo/pull/147137, and you should just wait for the patch to be forward-ported to 16.0.

Tisho99:
Thank you Pedro!

PR #193, opened 2024-01-04
[FIX] use lst_price for _compute_margin in product.product

anhvu-sg:
The margin calculation functions correctly for products without variants.
However, there is an issue with the calculation for products that have variants.

Screenshot at Jan 03 10-36-00

**Expected behavior**

Screenshot at Jan 03 10-35-39

### DIFF
--- a/product_standard_margin/models/product_product.py
+++ b/product_standard_margin/models/product_product.py
@@ -52,7 +52,7 @@ class ProductProduct(models.Model):
 
     # Compute Section
     @api.depends(
-        "list_price",
+        "lst_price",
         "product_tmpl_id.list_price",
         "standard_price",
         "taxes_id.price_include",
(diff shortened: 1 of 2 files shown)


product_standard_margin/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Product Margin and Margin Rate",
-    "version": "16.0.1.0.1",
+    "version": "16.0.1.0.2",
```
legalsylvain: This is unncessary. Will be done automatically, once merged.

legalsylvain (commented):
Hi @anhvu-sg . thanks a lot for sharing this bugfix.

Could you also add a test ? 

Thanks.

product_standard_margin/__manifest__.py
anhvu-sg: noted @legalsylvain

anhvu-sg:
> Could you also add a test ?

@legalsylvain, I've made the necessary updates. Could you please review it?

legalsylvain (approved):
Thanks !

dreispt (approved):

PR #198, opened 2024-03-08
[IMP] sale_margin[_delivered]_security: depend on product_cost_security

yajo:
Both modules do essentially the same thing: apply a restriction over product costs. Thus, it makes sense to share the same permission groups.

This refactor, that depends on https://github.com/OCA/product-attribute/pull/1538, improves the module readme and makes `sale_margin_security` auto-installable addon when `product_cost_security` is found.

A migration script is provided to make sure the same users still retain the same permissions.

---

[IMP] sale_margin_delivered_security: combine with product_cost_security

Following https://github.com/OCA/product-attribute/pull/1538, the security is now done via mixin and the view can be removed.

@moduon MT-5158


yajo:
I set as draft until https://github.com/OCA/product-attribute/pull/1538 is merged.

rafaelbn (approved):
In my side is a good improvement 👍🏼 

@legalsylvain , what is your opinion?

legalsylvain:
hi @rafaelbn, No point of view here. 
I don't use the ``_security`` OCA modules. In my company, product costs are open source, all people can read it ;-)

fcvalgar (approved):

Gelojr (approved):
Nice job @yajo LGTM.
https://www.loom.com/share/153c3c3e129047c38091273953c8b006?sid=8b5c3af3-51c0-4db8-bdfa-540ebb5cb31f
Maybe just add to the roadmap what I explain in the video about changing costs.

yajo:
Thanks @Gelojr. I added that in https://github.com/OCA/product-attribute/pull/1565 because that behavior comes from the lower-level module `product_cost_security`.

This one should be ready to merge IMHO.

PR #199, opened 2024-03-11
[16.0][IMP] product_margin_classification: set default company

aliciagaarzo:
Correctly transfer the default company and not the user's company.


amarcosg (approved):
Following multicompany Odoo [guidelines](https://www.odoo.com/documentation/16.0/developer/howtos/company.html#default-company) `self.env.company.id` is the correct way to avoid problems with multicompany users

legalsylvain (approved):
Thanks !

/ocabot merge patch

PR #200, opened 2024-03-19
[IMP] sale_report_margin  report field purchase_price

sistecem:
In the report it is more useful the total cost than the unit cost. 
It also gave wrong data when 2 lines of the same product where used

### DIFF
--- a/sale_report_margin/reports/sale_report.py
+++ b/sale_report_margin/reports/sale_report.py
@@ -8,15 +8,19 @@ class SaleReport(models.Model):
     _inherit = "sale.report"
 
     purchase_price = fields.Float(string="Purchase Price", readonly=True)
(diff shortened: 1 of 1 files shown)


sergio-teruel:
Sorry @sistecem The purchase_price field in Odoo core is by unit so to maintain data consistency it should be kept as a unit price.

Selección_2586

sistecem:
Thanks @sergio-teruel ,  I agree with keeping the consistency , but i think it should be AVG instead of SUM

image
image

sistecem:
I also added the field purchase_price_total , as an option in the report , resultin in something like this
image

sistecem:
Sorry for the mess, I am still learning to contribute to the OCA
I can make a new pull request if necessary 
Will do spanish translation if neccesary

PR #201, opened 2024-04-06
[14.0][FIX] account_invoice_margin: wrong default margin calculation

dreispt:
When installing the module, the Invoice total margin initial calculation
was including all AML lines, including taxes.
This resulted in a margin of more than 100%.

Example:
image


legalsylvain (approved):
Looks great !

marcelsavegnago (approved):

PR #202, opened 2024-04-06
[IMP] account_invoice_margin: init Invoice margins from SOs if possible

### LINKED ISSUE #182: account_invoice_margin: does not populate purchase_price on install
## Module
account_invoice_margin

## Describe the bug
When installing `account_invoice_margin`, the `purchase_price` field is expected to be populated with the product's `standard_price`, but it stays as zero.
Manually running `_compute_purchase_price` fixes it, but I can't see why this is not already being done on install - I could be missing something.

## To Reproduce
**Affected versions**: 14.0

1. Install account_invoice_margin
2. Find an Invoice using a product that has a non zero cost
3. Check the cost invoice line column: is is zero while it should be the product cost

**Expected behavior**
the cost invoice line column: is is zero while it should be the product cost

dreispt:
Includes #201 
Closes #182

### DIFF
--- a/account_invoice_margin/hooks.py
+++ b/account_invoice_margin/hooks.py
@@ -43,6 +43,7 @@ def pre_init_hook(cr):
             FROM account_move_line
             INNER JOIN account_move
             ON account_move.id = account_move_line.move_id
+            WHERE account_move_line.exclude_from_invoice_tab = false
             GROUP BY account_move_line.move_id
         )
         UPDATE account_move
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):

amarcosg (approved):

dreispt:
@sergio-teruel Would you like to merge?

sergio-teruel (approved):

PR #203, opened 2024-05-02
[16.0][ADD] product_margin_classification: Theoretical price actions on product list views

carmenbianca:
A fairly straightforward feature implementation.

Internal task: https://gestion.coopiteasy.be/web#id=11378&action=479&model=project.task&view_type=form&menu_id=536

### DIFF
--- a/product_margin_classification/models/product_product.py
+++ b/product_margin_classification/models/product_product.py
@@ -108,3 +108,16 @@ class Productproduct(models.Model):
     def use_theoretical_price(self):
         for product in self:
             product.lst_price = product.theoretical_price
+
+    def _apply_theoretical_price(self, state_list):
+        products = self.filtered(lambda x: x.margin_state in state_list)
+        products.use_theoretical_price()
+
+    def apply_theoretical_price(self):
+        self._apply_theoretical_price(["too_cheap", "too_expensive"])
+
+    def apply_theoretical_price_too_cheap(self):
+        self._apply_theoretical_price(["too_cheap"])
+
+    def apply_theoretical_price_too_expensive(self):
+        self._apply_theoretical_price(["too_expensive"])
--- a/product_margin_classification/views/view_product_product.xml
+++ b/product_margin_classification/views/view_product_product.xml
@@ -129,4 +129,33 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
         </field>
     </record>
 
+    <record id="action_apply_theoretical_price" model="ir.actions.server">
+        <field name="name">Apply Theoretical Price</field>
+        <field name="groups_id" eval="[Command.link(ref('base.group_user'))]" />
+        <field name="model_id" ref="product.model_product_product" />
+        <field name="binding_model_id" ref="product.model_product_product" />
+        <field name="binding_view_types">list</field>
+        <field name="state">code</field>
+        <field name="code">records.apply_theoretical_price()</field>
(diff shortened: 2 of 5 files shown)


product_margin_classification/views/view_product_product.xml
```diff
@@ -129,4 +129,33 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
         </field>
     </record>
 
+    <record id="action_apply_theoretical_price" model="ir.actions.server">
+        <field name="name">Apply Theoretical Price</field>
+        <field name="groups_id" eval="[(4, ref('base.group_user'))]"/>
```
carmenbianca: Here and elsewhere, I am not certain this is the correct value for groups_id.

product_margin_classification/models/product_product.py
```diff
@@ -108,3 +108,18 @@ def _get_margin_info(
...
+    def _apply_theoretical_price(self, state_list):
+        products = self.filtered(
+            lambda x: x.margin_state in state_list
+        )
+        products.use_theoretical_price()
+
+    def apply_theoretical_price(self):
+        self._apply_theoretical_price(["too_cheap", "too_expensive"])
+
+    def action_apply_theoretical_price_too_cheap(self):
```
legalsylvain:
```suggestion
    def apply_theoretical_price_too_cheap(self):
```

product_margin_classification/models/product_product.py
```diff
@@ -108,3 +108,18 @@ def _get_margin_info(
...
+        self._apply_theoretical_price(["too_cheap"])
+
+    def action_apply_theoretical_price_too_expensive(self):
```
legalsylvain:
```suggestion
    def apply_theoretical_price_too_expensive(self):
```

product_margin_classification/views/view_product_product.xml
legalsylvain: the function will change the sale price for many products. 
it should be available for people that can do that operation for a product.

legalsylvain (commented):
Some fixes inline.

Could you add simple test for new functions ? 

Otherwise, LGTM. thanks !

product_margin_classification/views/view_product_product.xml
carmenbianca: Changed this to `base.group_system`, which appears to be the group required for writing to product.product.

carmenbianca:
Added tests for the added actions.

Did not add tests for directly executing the methods. It would be a bit of a test-the-same-thing-twice affair. But if needed, I can add them for completeness' sake.

product_margin_classification/views/view_product_product.xml
legalsylvain: That's a problem. base.group_system is erp manager. The feature will be hidden to all the users. To fix it : 
- set ``base.group_user``

(Core code : 
```access_product_product_employee,product.product employee,model_product_product,base.group_user,1,0,0,0```)

so the feature will be visible for all the user that can see products.
Then an  error will be raised by ORM, if user can not write on products.

legalsylvain (changes requested):
Thanks. A remark inline regarding incorrect group. otherwise, LGTM !

product_margin_classification/views/view_product_product.xml
carmenbianca: reverted, fixed :+1:

PR #205, opened 2024-05-09
 [16.0][ADD] stock_picking_product_link_margin

carmenbianca:
Super simple module. The customer need is to more easily access the margin list view from stock pickings.

Internal task: https://gestion.coopiteasy.be/web#id=11376&action=479&model=project.task&view_type=form&menu_id=536


victor-champonnois:
To give a bit more context : this module is for supermarkets (or any store) who use `product_margin_classification`  but want to apply the margin when receiving the delivery, rather than when receiving the invoice.

victor-champonnois (approved):
Functionnal review OK

carmenbianca:
@victor-champonnois added your target audience description to the documentation

carmenbianca:
Not stale :) @remytms can you review?

carmenbianca:
Hi @legalsylvain. I've added a note to `CONFIGURE.rst`:

> In order to use this module, you must enable product variants. The necessary margin view does not exist for product templates.

But our client doesn't want to enable product variants, so I'll have to chase this down.

Before I run off to apply patches everywhere, is there a reason why the list view is for `product.product` instead of `product.template`?

PR #206, opened 2024-05-24
[16.0][IMP] product_margin_classification: modify registration rule to solve multi-company error

aliciagaarzo:
Working with classification margins, when we have a multi-company user when changing companies and trying to create a classification margin the following error occurs:

error

If the domain of the registration rule is modified, the problem is solved

### DIFF
--- a/product_margin_classification/security/ir_rule.xml
+++ b/product_margin_classification/security/ir_rule.xml
@@ -13,9 +13,7 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
             ref="product_margin_classification.model_product_margin_classification"
         />
         <field name="global" eval="True" />
-        <field
-            name="domain_force"
-        >['|', ('company_id', '=', user.company_id.id), ('company_id','=',False)]</field>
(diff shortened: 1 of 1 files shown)


product_margin_classification/security/ir_rule.xml
```diff
@@ -15,7 +15,7 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
         <field name="global" eval="True" />
         <field
             name="domain_force"
-        >['|', ('company_id', '=', user.company_id.id), ('company_id','=',False)]</field>
+        >['|', ('company_id', 'in', company_ids), ('company_id', '=', False)]</field>
```
legalsylvain:
```suggestion
        >[('company_id', 'in', company_ids + [False])]</field>
```

See performance improvment explained here : https://github.com/odoo/odoo/pull/157861/files

legalsylvain (changes requested):
Hi @aliciagaarzo. Thanks for proposing your patch.

a minor improvment proposed inline.

otherwise, LGTM. Thanks !

legalsylvain:
Thanks ! 

/ocabot merge patch

PR #207, opened 2024-05-31
[FIX] [16.0] sale_margin_delivered: Fix calculation when more quantities are sent than ordered

Shide:
When sending more quantities than ordered, `purchase_price_delivery` field is not calculated correctly in this Odoo version.

Now, instead of use unit_cost of valuation layer, value / quantity is used instead.
Refunds are already taken in consideration in valuation layers.
Before, only the first valuation layer is taken in consideration, now all valuation layers are considered. (FIFO product valuation and one move needs to be splitted in 2 stock.move.lines and each move line has different lots with different values)

Also added tests, more concise translations and "help" to fields to know exactly what does.

https://www.loom.com/share/9cc8c826b8054b62a508d561632b837c?sid=ce29f0dd-9fd4-40bd-b615-4f775f42b339

MT-6136 @moduon @yajo @EmilioPascual  @rafaelbn @sergio-teruel please review if you want :)

### DIFF
--- a/sale_margin_delivered/README.rst
+++ b/sale_margin_delivered/README.rst
@@ -92,6 +92,9 @@ Contributors
   * Carlos Roca
   * Pilar Vargas
 
+* Eduardo de Miguel (`Moduon <https://www.moduon.team/>`__)
+* Rafael Blasco (`Moduon <https://www.moduon.team/>`__)
+
 Maintainers
 ~~~~~~~~~~~
 
--- a/sale_margin_delivered/i18n/es.po
+++ b/sale_margin_delivered/i18n/es.po
@@ -4,34 +4,45 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 11.0\n"
+"Project-Id-Version: Odoo Server 16.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-12-09 08:52+0000\n"
-"PO-Revision-Date: 2023-10-09 09:14+0000\n"
-"Last-Translator: Ivorra78 <informatica@totmaterial.es>\n"
+"POT-Creation-Date: 2024-05-31 07:23+0000\n"
+"PO-Revision-Date: 2024-05-31 09:30+0200\n"
+"Last-Translator: \n"
 "Language-Team: \n"
 "Language: es\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=2; plural=n != 1;\n"
-"X-Generator: Weblate 4.17\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"X-Generator: Poedit 3.4.4\n"
+
+#. module: sale_margin_delivered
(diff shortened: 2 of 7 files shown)


yajo (approved):

Gelojr (approved):
Nice @Shide LGTM

sergio-teruel (approved):

yajo:
Could you merge please @sergio-teruel? I don't have permission here

sergio-teruel:
Ohhh!! I'm not a PSC... :sob:

rafaelbn:
Hello all! @sergio-teruel @Gelojr @yajo @Gelojr !

I've reviewed and I have doubts in cost price delivered and a better order in columns 😇 

This is my video review (1.5x is 5 minutes)

https://www.loom.com/share/3ffc172466d24629bee5247cdb89f9f6?sid=f0146094-5bf3-4798-936f-5ea6bc32810e

🙏🏼 Thank you! Reviewing with ❤️ 😄 

imagen

Gelojr:
> Hello all! @sergio-teruel @Gelojr @yajo @Gelojr !
> 
> I've reviewed and I have doubts in cost price delivered and a better order in columns 😇
> 
> This is my video review (1.5x is 5 minutes)
> 
> https://www.loom.com/share/3ffc172466d24629bee5247cdb89f9f6?sid=f0146094-5bf3-4798-936f-5ea6bc32810e
> 
> 🙏🏼 Thank you! Reviewing with ❤️ 😄
> 
> imagen

I believe that the Cost price dlvd. column should appear, since it is the total cost of what is delivered. It seems that it is not calculated correctly, since it is calculating the cost of the quantity ordered.

Shide:
> Hello all! @sergio-teruel @Gelojr @yajo @Gelojr !
> 
> I've reviewed and I have doubts in cost price delivered and a better order in columns 😇
> 
> This is my video review (1.5x is 5 minutes)
> 
> https://www.loom.com/share/3ffc172466d24629bee5247cdb89f9f6?sid=f0146094-5bf3-4798-936f-5ea6bc32810e
> 
> 🙏🏼 Thank you! Reviewing with ❤️ 😄
> 
> imagen

Those columns are not duplicated, but in your case, you are using Standard or AVCO product valuation.
Those columns are not meaningfull in your context, but in a FIFO valuation should change a bit. 

In any case, for the commercial user, Cost and Margin are welcome because they cannot see dlvd. margins when doing the quotation. 
If commercial user receives a % extra payment for each sale based on the margin (assume margin dlvd. because cost can be modified freely on the line) and valuations are Standard or AVCO, they can assume that dlvd. cost and margins are going to be the same as sale cost and margins.

Then,
- for commercial users: hide dlvd. columns and not modify cost in the sale line (because they are not going to receive more %).
- for reporter users: hide cost and margin columns and show dlvd. columns because they need to see the real value of the delivered quantities to calculate % extra payments.

Since all columns can be used for different type of users and Odoo allows to custom tree columns per user, the code as it is will work for a variety of roles. 
I won't add anything more. Just inform users that what do each column and what are suggested columns to see on each role.

rafaelbn:
After discussing with @Shide , this ho ahead as resolves BUG and later we could improve it.

Thank you!! 😃

rafaelbn (approved):
/ocabot merge patch

PR #210, opened 2024-06-12
[IMP] product_standard_margin : Remove 'theoretical' in the name of field

legalsylvain:
- as it is based on the field cost and sale price, that doesn't contain 'theoretical'
- save place in form / tree view.


quentinDupont (approved):
makes sense

PR #213, opened 2024-06-18
[ADD] [16.0] sale_margin_delivered_dropshipping: Calculate Dropship moves too

Shide:
Calculate costs for Dropship moves too and refunds of Dropship.

Also added myself as a maintainer.

PR Series: 
- https://github.com/OCA/margin-analysis/pull/213 (this one)
- https://github.com/OCA/sale-reporting/pull/275

MT-6512 @moduon @rafaelbn @fcvalgar @Gelojr please review if you want :)


sale_margin_delivered/models/sale_margin.py
```diff
@@ -53,20 +53,28 @@ def _compute_margin_delivered(self):
...
-                moves = line.move_ids.filtered(
-                    lambda x: (
-                        x.state == "done"
-                        and (
-                            x.picking_code == "outgoing"
-                            or (x.picking_code == "incoming" and x.to_refund)
+                valuation_layers = self.env["stock.valuation.layer"]
+                for move in line.move_ids.filtered(lambda m: m.state == "done"):
+                    if move.picking_code == "outgoing":
+                        # Outgoing moves has 1 valuation layer and are always negative
```
yajo:
```suggestion
                        # Outgoing moves have 1 valuation layer and are always negative
```

sale_margin_delivered/models/sale_margin.py
```diff
@@ -53,20 +53,28 @@ def _compute_margin_delivered(self):
...
+                        valuation_layers |= move.stock_valuation_layer_ids
+                    elif move.picking_code == "incoming" and move.to_refund:
+                        # Incoming moves has 2 valuation layers. Use positive one
```
yajo:
```suggestion
                        # Incoming moves have 2 valuation layers. Use positive one
```

sale_margin_delivered/models/sale_margin.py
```diff
@@ -53,20 +53,28 @@ def _compute_margin_delivered(self):
...
                         )
-                    )
-                )
-                # Qty and Value Delivered is negative when outgoing
-                value_delivered = sum(moves.mapped("stock_valuation_layer_ids.value"))
+                    elif (
+                        move.location_dest_usage == "customer"
+                        and move.location_usage == "supplier"
+                    ):
+                        # Dropship moves has 2 valuation layers. Use negative one
```
yajo:
```suggestion
                        # Dropship moves have 2 valuation layers. Use negative one
```

yajo (approved):

Gelojr (approved):
Nice Job @Shide LGTM
https://www.loom.com/share/33e5b02b50534387b78cfe5b954c4a93?sid=42f717a1-c1a2-4085-9442-b6a9f50a7b98

fcvalgar (approved):
Function review. 

LGTM, thank you @Shide

lmignon (changes requested):
@Shide Thank you for this fix. Can you add some tests to cover these cases.

Shide:
@lmignon Done.

I've tryied to not depends from stock_dropshipping but I can't simulate at 100% the same behavior as stock_dropshipping does (was not taking well the returned moves of dropshippings)

So, this is why I've splitted the logic in 2 commits:

- REF: Leave as it is, but prepared to add value layers (for dropshipping) + Improved README
- ADD: The new dropshipping module with its tests. (And easy to migrate)

Shide:
I want you all to review again the modules because has been splitted in 2.

sale_margin_delivered/models/sale_margin.py
```diff
@@ -31,6 +31,20 @@ class SaleOrderLine(models.Model):
         "Formula: Value Delivered / Quantity Delivered",
     )
 
+    def _get_delivered_margin_valuation_layers(self):
```
lmignon, line 42: @Shide we could opt for a more meaningful method name
```suggestion
    def _get_delivered_qty_and_value(self):
        """ Return the qty and the value of delivered products
        """
        ....
```

What do you think of also returning the qty delivered via this method. This way, we remove the use of valuation layers in the '_compute_margin_delivered' method and we make-it plug-gable...

Gelojr (approved):
Nice @Shide LGTM

https://www.loom.com/share/ef145bc1d7d24a8098c32aea691b8b21?sid=d5ea3bd0-2e1e-4adb-861c-bc3854b9979c

sale_margin_delivered/models/sale_margin.py
Shide: The quantity and value fields are always calculated the same based on the valuation_layers (sum).

Having the function that decides which Valuation Layers are going to be used, allows us to abstract from the rest of the flow and to be able to easily inherit in the future if we decide to add another use case (as I have done with Dropshipping).

If I added the Value and Quantity to the function, every time we inherited we would have to reselect all the Valuation Layers involved in calculation, and that would not be possible if there are several modules inheriting the same function.

TLDR; I have decided not to put it in the function because the calculation of the quantity and value is always the same and would not allow a simple inheritance.

I'm going to improve doc of the functions

lmignon (approved):

fcvalgar (approved):

PR #214, opened 2024-06-18
[16.0][FIX] sale_margin_delivered: Avoid rounding error

lmignon:
'price_reduce' is deprecated and removed into the next version. Compute
the prirce_reduct from the price_subotal / product_uom_qty. We might
be tempted to use the 'price_reduce_taxecl' field from the sale order
line but this field is rounded by default to the monetary precision.

As an additional benefit this change ensures the compatibility with
the 'sale_triple_discount' addon. Indeed, when
'sale_triple_discount' is installed, the discount field is not used
as an aggregation of all the applied discount. It's only use to store
the first discount applied. Therefore, the field is not properly
computed since it doesn't include the second and third discount.

### DIFF
--- a/sale_margin_delivered/models/sale_margin.py
+++ b/sale_margin_delivered/models/sale_margin.py
@@ -72,6 +72,10 @@ class SaleOrderLine(models.Model):
         self.margin_delivered_percent = 0.0
         self.purchase_price_delivery = 0.0
         for line in self.filtered("qty_delivered"):
+            # we need to compute the price reduce to avoid rounding issues
+            # the one stored in the line is rounded to the product price precision
+            price_reduce_taxexcl = line.price_subtotal / line.product_uom_qty
+
             if line.product_id.type != "product":
                 currency = line.order_id.pricelist_id.currency_id
                 price = line.purchase_price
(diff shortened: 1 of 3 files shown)


lmignon:
@Shide @yajo Can you take a look at this one plz?

Shide (approved):
Wonderful, checked code and does the same

lmignon:
@shide I'll change this since `price_reduce_taxexcl` is rounded and therefore the margin is not properly computed when we've more than 1 product.

lmignon:
@Shide @yajo @Gelojr @fcvalgar In Odoo standard margin in % is expressed as a decimal... 1 = 100%, 0.5=50% ... In this module margin % is expressed as 100 = 1000%, 50 = 50%... what's the motivation. It's strange to store precentage value in a different way than Odoo. Can I change this and provide a migration script?

fcvalgar (changes requested):
I can't see the percentage symbol in the delivered margin.

image

fcvalgar (approved):

Shide:
> @Shide @yajo @Gelojr @fcvalgar In Odoo standard margin in % is expressed as a decimal... 1 = 100%, 0.5=50% ... In this module margin % is expressed as 100 = 1000%, 50 = 50%... what's the motivation. It's strange to store precentage value in a different way than Odoo. Can I change this and provide a migration script? 

Sure!

rafaelbn:
> @Shide @yajo @Gelojr @fcvalgar In Odoo standard margin in % is expressed as a decimal... 1 = 100%, 0.5=50% ... In this module margin % is expressed as 100 = 1000%, 50 = 50%... what's the motivation. It's strange to store precentage value in a different way than Odoo. Can I change this and provide a migration script?

This would be an amazing great contribution @lmignon ! 💪🏼 😄 Thank you!

rousseldenis (approved):
Code review

Shide:
Could be closed in favor of https://github.com/OCA/margin-analysis/pull/215

lmignon:
included into #215

PR #215, opened 2024-06-19
[16.0] sale_margin_delivery: Store percentage value as into odoo

lmignon:
**Prior to this change**

* margin_percent
  100% = 1.0
  50% = 0.5

* margin_delivered_percent
  100%  = 100
  50% = 50

**After**
* margin_percent
  100% = 1.0
  50% = 0.5

* margin_delivered_percent
  100%  = 1.0
  50% = 0.5

* [ ] includes #214

### DIFF
--- a/sale_margin_delivered/migrations/16.0.2.0.0/post-migration.py
+++ b/sale_margin_delivered/migrations/16.0.2.0.0/post-migration.py
@@ -1,0 +1,15 @@
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version=None):
+    """Store margin delivered precentage as a fraction of 1"""
+    cr.execute(
+        """
+        UPDATE sale_order_line
+        SET margin_delivered_percent = margin_delivered_percent / 100
+        WHERE margin_delivered_percent > 0
+        """
+    )
+    _logger.info("Updated %d records", cr.rowcount)
--- a/sale_margin_delivered/__manifest__.py
+++ b/sale_margin_delivered/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Sale Margin Delivered",
-    "version": "16.0.1.0.5",
+    "version": "16.0.2.0.0",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/margin-analysis",
     "category": "Sales",
--- a/sale_margin_delivered/models/sale_margin.py
+++ b/sale_margin_delivered/models/sale_margin.py
@@ -74,7 +74,11 @@ class SaleOrderLine(models.Model):
         for line in self.filtered("qty_delivered"):
             # we need to compute the price reduce to avoid rounding issues
             # the one stored in the line is rounded to the product price precision
-            price_reduce_taxexcl = line.price_subtotal / line.product_uom_qty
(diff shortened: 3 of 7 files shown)


lmignon:
@Shide @yajo @Gelojr @fcvalgar  following my comment https://github.com/oca/margin-analysis/pull/214#issuecomment-2178844145

fcvalgar (approved):
LGTM, sorry I dont see this PR.

Thank you @lmignon

rousseldenis:
@lmignon It seems dropshipping module should be updated too.

sale_margin_delivered/migrations/16.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(env, version):
+    """Store margin delivered precentage as a fraction of 1"""
+    env.cr.execute(
+        """
+        UPDATE sale_order_line
+        SET margin_delivered_percent = margin_delivered_percent / 100
```
pedrobaeza, line 11: Should this be rounded?

lmignon: @pedrobaeza This field is a float field without precision so it should not be rounded.

pedrobaeza: But AFAIK, when no rounding is applied, web client applies 2 digits rounding when saving from it. Is this field shown in web client and thus stripped by it?

lmignon:
> @lmignon It seems dropshipping module should be updated too.

it's done.

sale_margin_delivered/migrations/16.0.2.0.0/post-migration.py
pedrobaeza: In a quick test in runboat, it seems that at least 4 digits are respected, so this seems OK.

lmignon: rounding is done on the client side by the use of the 'percentage' widget

pedrobaeza: Yeah, my fear was that on new records, a rounding would be also applied server-side, and then, data was not consistent on analysis. I remember some time ago to have this problem (reported by @alexis-via) with `account.move.line` with `credit` and `debit` not rounded to 2 digits, and this causing differences in aggregations.

lmignon: If no precision is specified on the field declaration, values are never rounded on server side. It's required to avoid rounding issues as the sum of rounded values != to the round of the sum of values.

pedrobaeza: I didn't express correctly, sorry. It was not server-side, but the web client passing rounded values to the server. Anyway, this seems clear and tried on a functional test. Thanks for the clarifications.

pedrobaeza: I have a mix of both kind of issues (server side and client side) in my memories, and I want to avoid both.

yajo (approved):

Shide (approved):
Thanks for the awesome update! ❤️

Shide:
/ocabot merge nobump

Done with nobump because version is already updated manually

PR #216, opened 2024-07-08
[16.0][IMP] account_invoice_margin_sale: Add post_install hook

carolinafernandez-tecnativa:
- Add post_init_hook to update purchase cost in invoice lines from sale order when installing module.

@Tecnativa
TT49994
@pedrobaeza @victoralmau

### DIFF
--- a/account_invoice_margin_sale/hooks.py
+++ b/account_invoice_margin_sale/hooks.py
@@ -1,0 +1,39 @@
+def post_init_hook(cr, registry):
+    # update purchase_price for invoices from sale_order
+    cr.execute(
+        """
+        UPDATE account_move_line AS aml
+        SET
+            purchase_price = sol.purchase_price,
+            margin = sol.margin,
+            margin_signed = CASE
+                WHEN am.move_type = 'out_refund' THEN sol.margin * -1
+                ELSE sol.margin
+            END,
+            margin_percent = sol.margin_percent * 100
+        FROM sale_order_line sol, account_move am, sale_order_line_invoice_rel rel
+        WHERE am.id = aml.move_id
+            AND rel.order_line_id = sol.id
+            AND rel.invoice_line_id = aml.id;
+        """
+    )
+    # recalculate margin for invoices from sale_order
+    cr.execute(
+        """
+        UPDATE account_move AS am
+        SET
+            margin = aml.sum_margin,
+            margin_signed = aml.sum_margin_signed,
+            margin_percent = aml.sum_margin / am.amount_untaxed * 100
+        FROM (
+            SELECT
+                aml.move_id,
+                SUM(aml.margin) AS sum_margin,
+                SUM(aml.margin_signed) AS sum_margin_signed
+            FROM account_move_line AS aml
+            GROUP BY aml.move_id
+        ) AS aml
+        WHERE am.id = aml.move_id
+        AND am.amount_untaxed > 0.0;
+        """
+    )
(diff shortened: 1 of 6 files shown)


account_invoice_margin_sale/hooks.py
```diff
@@ -0,0 +1,17 @@
+from odoo import SUPERUSER_ID, api
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    sale_orders = env["sale.order"].search([("invoice_ids", "!=", False)])
```
pedrobaeza: This must be done by SQL, as this is totally inefficient.

account_invoice_margin_sale/hooks.py
```diff
@@ -0,0 +1,51 @@
...
+            ON soi.order_line_id = order_line.id
+            WHERE soi.invoice_line_id = aml.id
+        )
+        WHERE EXISTS (
+            SELECT 1
+            FROM sale_order_line AS order_line
+            INNER JOIN sale_order_line_invoice_rel AS soi
+            ON soi.order_line_id = order_line.id
+            WHERE soi.invoice_line_id = aml.id
+        );
```
pedrobaeza: Please don't complicate SQL:

```suggestion
        UPDATE account_move_line AS aml
        SET purchase_price = sol.purchase_price,
            margin = ...
        FROM sale_order_line sol
        INNER JOIN sale_order_line_invoice_rel AS rel
        ON rel.order_line_id = sol.id
        WHERE rel.invoice_line_id = aml.id
```

and do something similar for the other one.

account_invoice_margin_sale/hooks.py
```diff
@@ -0,0 +1,37 @@
+def pre_init_hook(cr, registry):
+    # update purchase_price for invoices from sale_order
+    cr.execute(
+        """
+        UPDATE account_move_line AS aml
+        SET purchase_price = (
+            SELECT COALESCE(SUM(sol.purchase_price), 0.0)
```
pedrobaeza: This is not what I have suggested. If mine is not 100% correct, you should fix that one, not to put this one that is totally incorrect and if OK, totally inefficient.

carolinafernandez-tecnativa: could you please check now?

pedrobaeza: Yes, but it seems the same SQL I put before. Why not putting it in the first place? And what about the margin field and the second SQL should be optimized as well.

carolinafernandez-tecnativa: should we improve this as well? https://github.com/OCA/margin-analysis/blob/16.0/account_invoice_margin/hooks.py#L36

carolinafernandez-tecnativa:
ping @pedrobaeza

account_invoice_margin_sale/hooks.py
```diff
@@ -0,0 +1,35 @@
...
+            margin = aml.sum_margin,
+            margin_signed = aml.sum_margin_signed,
+            margin_percent = aml.sum_margin_signed / am.amount_untaxed * 100
+        FROM (
+            SELECT
+                aml.move_id,
+                SUM(aml.margin) AS sum_margin,
+                SUM(aml.margin_signed) AS sum_margin_signed
+            FROM account_move_line AS aml
+            INNER JOIN sale_order_line_invoice_rel AS rel ON rel.invoice_line_id = aml.id
```
pedrobaeza: This JOIN is doing nothing. Please remove it.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #218, opened 2024-08-29
[17.0] [MIG] product margin classification: Migration to 17.0

SodexisTeam:


legalsylvain:
the module is marked as installable false. could you set the PR in draft state and ping when it's ready ? thanks

.pre-commit-config.yaml
```diff
@@ -1,6 +1,7 @@
 exclude: |
   (?x)
   # NOT INSTALLABLE ADDONS
+  ^product_margin_classification/|
```
legalsylvain, line 4: Why this change ?

legalsylvain (changes requested):

Issue #222, opened 2024-12-11
sale_margin_security: Don't depend on product cost security


rousseldenis:
People that have access to cost fields should not have access necessarily to sale margin ones.

We should restore the previous implementation.

Due to : https://github.com/OCA/margin-analysis/pull/198

@yajo @rafaelbn @lmignon @phschmidt

yajo:
If you know the sale price and the margin, you know the cost... All you have to do is a simple multiplication. And vice-versa: if you know the cost and the sale price, with simple math you get the margin.

So, IIUC, both things are the same, and that's why it was refactored for simplification.

rousseldenis:
> If you know the sale price and the margin, you know the cost... All you have to do is a simple multiplication. And vice-versa: if you know the cost and the sale price, with simple math you get the margin.
> 
> So, IIUC, both things are the same, and that's why it was refactored for simplification.

Yes, but some companies want to split the access rights to explicit margin fields visibility (there are buyers and salesmen).

Of course, people can compute it but they will need to do the exercise.

Restablishing the previous behavior will allow to be flexible and do both scenarios.

yajo:
> they will need to do the exercise

... and they'll do it. Trust me, it happened. I'm sorry to say this, but I disagree with this proposal. 😅 

These modules are labeled "security" for a reason. Restoring the previous behavior would be useful only to reduce security and UX. You would install the modules, but you would have an option to only get a placebo-like sense of security regarding cost information. It'd be a step backwards.

If you still need to do that, I suggest that you do it in a separate module that depends on these ones. I don't think this behavior should be supported by the base ones.

rousseldenis:
I don't really want to restore previous behavior:

- Let the cost group on `purchase_price` field.
- Restore the `group_sale_margin`on margin fields.
- Add implied_ids on margin group to cost group

- Maybe add a mixin for margin as for cost

rousseldenis:
@jdidderen @jdidderen-nsi  As you did the migration for v17

What do you think ?

PR #223, opened 2024-12-11
[IMP] sale_margin_security: Allows more flexibility in groups

rousseldenis:
As in some flows, people want to split fields visibility to buyers and salesmen, restore the margin group. As in cost module, there is two groups, add a group for sale margins + edit costs


lmignon (approved):
Granular security is easier to manage.... Thank you @rousseldenis

sergio-teruel (approved):

rousseldenis:
@OCA/accounting-maintainers Could you link this to :

https://github.com/OCA/margin-analysis/issues/222

Thanks

yajo (changes requested):
In https://github.com/OCA/margin-analysis/issues/222, we already touched upon this matter. From my perspective, this approach seems to undermine the intended security, as it effectively turns the security into a placebo. Anyone with access to margin will essentially have access to costs, which, in practical terms, means the same thing.

In the previous PRs from https://github.com/OCA/product-attribute/pull/1538#issue-2171562830 and https://github.com/OCA/margin-analysis/pull/198#issue-2175629776, we addressed the inconsistencies and security issues that made the previous status difficult to manage effectively. Thanks to those contributions, the UX has significantly improved, and we now have a simple, intuitive permissions dropdown:

image

This solution made security clear and functional right out of the box, which I believe was a positive step forward.

If we were to merge this new approach, however, we would return to the need for debug mode to configure security permissions, which could lead to inconsistent states. For example, here are some instances from Runboat:

image

image

By contrast, I think the thorough functional validation done by @Gelojr in https://github.com/OCA/margin-analysis/pull/198#pullrequestreview-1951828542 demonstrates how the current approach ensures consistent behavior across the board.

That said, I still can see that adding an intermediate permission option to the dropdown could be a valuable enhancement, and it could look like this:

1. Read.
2. Edit only in SO.
3. Edit.

I understand there could be use cases where you might want salespeople to edit costs in the SO but not in the product. However, beyond this point, I believe adding more complexity would only lead to confusion and potential misconfigurations, without genuinely improving the system. For that reason, I don't agree with the proposed change in its current form.

In the end, my concern is that increasing complexity may not make the system easier to manage, but rather more prone to mistakes and misconfigurations. 


MT-8486

rafaelbn (commented):
Hello! 😄 

We would appreciate your concerns about maintaining flexibility without adding unnecessary complexity or creating a false sense of security. The key challenge is to offer more granular visibility options without making the configuration confusing or difficult to manage.

How do you think we could achieve this balance? For example, introduce the new permission level that give teams the control they need—without requiring debug mode or hidden configurations—while still keeping the user experience clear and straightforward?

The tooltip could be clear to explain the groups so our users don't mistake with them.

I’d really value your input on how to simplify and adapt these settings to better serve everyone’s needs.

I agree with @yajo , and also I can understand that something that for us our our clients don't have any sense for other have it. The goal is to fit everyone.

Thank you! ❤️

rousseldenis:
Hi,

I didn't see the category in groups configuration.

I will add the category and configuration will show:

image

rousseldenis:
@rafaelbn If I understand, your main concern is about user interface. And I understand it. We usually minimize the impact of configuration at that level (using base_user_role module or letting the administrator manages it by itself). :sweat_smile:

rousseldenis:
> I understand there could be use cases where you might want salespeople to edit costs in the SO but not in the product.

That's not our use case. The reason of that change is about margin visibility (and I know costs people can compute them).

People that could have access to costs could not necessarily have access (direct) to margins.

rousseldenis:
@rafaelbn The problem to dsiplay a selection is that the groups are not strictly hierarchical.

yajo:
Indeed. Let's try to get to something more usable and sensible.

Only the costs fields are actually editable, because margins are computed, right? So, the edit mode on margins would make no sense IMHO.

Option A, if you want to split displaying or hiding margins, is to divide between a dropdown and a checkbox. Something like this:

Diagrama sin título drawio


However, IMHO it also doesn't make sense to let somebody actually *edit* the costs without being able to *read* the margins. If you have edit permissions, then I can imagine that you need to know the margin to be able to edit the cost appropriately, right? If so, this Option B would fit better:

Diagrama sin título drawio (1)


As you can see, both designs feature a warning when the user is creating an inconsistency. This module is suffixed `_security` for a reason, and by allowing only one of the 2 read levels, you are only hiding the information, not blocking it. Thus it's important that the admin gets that warning clearly displayed.

WDYT?

ghost (approved):
I understand the changes and i don't think it hurts anyone to have it.
I agree in principle that having access to the cost gives an idea of the margin. 
However, I don't see the issue with the approach proposed by Denis, as the old approach is still possible and the two can co-exist.

For the UX debate, can we simply have the following ?
image
image
image

yajo:
IIUC https://github.com/OCA/margin-analysis/pull/223#pullrequestreview-2527818423, your proposal is 4 levels of permissions, each one inheriting the previous one:
1. Nothing.
2. Read costs.
3. Read costs and margin.
4. Edit costs and margin.

Is that correct? Would that be good for you @rousseldenis?

ghost:
> IIUC [#223 (review)](https://github.com/OCA/margin-analysis/pull/223#pullrequestreview-2527818423), your proposal is 4 levels of permissions, each one inheriting the previous one:
> 
> 1. Nothing.
> 2. Read costs.
> 3. Read costs and margin.
> 4. Edit costs and margin.
> 
> Is that correct? Would that be good for you @rousseldenis?

Module categories : 
- Cost Security (new one)
   - Product Costs (existing one)
   - Sale Margins (new one)

Security groups :
- Product Costs
   - Read (existing one)
   - Edit (existing one)
- Sale Margins
   - Show sale margin (new one)

And the new group has a inherit (implied_ids) on the read security group from product costs. 

The edit security group from product costs is isolated and managed manually. 

You can check it on the runboat linked to the PR.

yajo:
Could you please explain to me why would somebody be able to edit costs and yet forbidden to see the margin? Isn't that essential information while editing costs? 🤔

rafaelbn:
Hello all,

What I talked with @rousseldenis in a Google Meet was this result:

imagen

Best regards 😄 🙏🏼 ❤️ 

@moduon MT-8486

rousseldenis:
> Hello all,
> 
> What I talked with @rousseldenis in a Google Meet was this result:
> 
> imagen
> 
> Best regards 😄 🙏🏼 ❤️
> 
> @moduon MT-8486

I finally reverted as the yellow line case is not possible as groups should imply the preceding to be shown as select box.

 @yajo @rafaelbn IMHO, if you want to get the margin access and costs in one operation, you should think about using `base_user_role` module.

rousseldenis:
At this time, the groups are managed like this:

image

like suggested

lmignon (approved):

sale_margin_security/models/sale_order.py
```diff
@@ -6,16 +6,20 @@
 
 class SaleOrder(models.Model):
     _name = "sale.order"
-    _inherit = ["sale.order", "product.cost.security.mixin"]
+    _inherit = ["sale.order"]
```
yajo, line 9: issue: undo this change please. Otherwise, users with read-only permissions for cost will actually be able to write the fields.

yajo (changes requested):
Thanks for the update!

I see that you're finally implementing option A from https://github.com/OCA/margin-analysis/pull/223#issuecomment-2556487578.

I still miss the warning when somebody selects read-only permissions only on either costs or margins. We should make it clear to the sysadmin that the user can compute the other thing easily with a calculator. We're managing user security here, and even if we might trick the user, we shouldn't trick the sysadmin. As a very minimum fix, it should be in the readme.

There's also a bug IIUC, noted below.

sale_margin_security/models/sale_order.py
rousseldenis: I'll check that

rousseldenis: @yajo In this case, sale order has no field that should be protected by cost group, correct? Only lines should (as it is still now) for `purchase_price`

yajo: Ah! you're totally right.

PR #225, opened 2025-01-08
[FIX] product_margin_classification: compute correctly fields in multi company context.

legalsylvain:
new computed and stored fields depends on standard_price that is company dependant. As a result, add a warning in the ROADMAP section, to explicit current module limitation. Ensure that computation is done in the context of the company of the product, to avoid to base computation on bad / null standard prices

### DIFF
--- a/product_margin_classification/README.rst
+++ b/product_margin_classification/README.rst
@@ -91,6 +91,11 @@ Known issues / Roadmap
   due to the poor design of Odoo product module.
   This issue can be maybe fixed in new version of Odoo.
 
+* This module will not work for products that available on many
+  companies, because this module introduces stored fields, that can
+  not be company dependant, and the fields depends on ``standard_price``
+  that is company dependant.
+
 Bug Tracker
 ===========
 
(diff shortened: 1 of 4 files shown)


product_margin_classification/README.rst
```diff
@@ -91,6 +91,11 @@ Known issues / Roadmap
   due to the poor design of Odoo product module.
   This issue can be maybe fixed in new version of Odoo.
 
+* This module will not work for products that available on many
```
rvalyi, line 94: s/that available/that are available/
s/on many/in several/

product_margin_classification/README.rst
```diff
@@ -91,6 +91,11 @@ Known issues / Roadmap
...
+  companies, because this module introduces stored fields, that can
+  not be company dependant, and the fields depends on ``standard_price``
```
rvalyi, line 96: s/can not/cannot/
s/dependant/dependent/
s/depends/depend/ (plural)

product_margin_classification/README.rst
```diff
@@ -91,6 +91,11 @@ Known issues / Roadmap
...
+  that is company dependant.
```
rvalyi, line 97: s/dependant/dependent/

rvalyi (commented):
LGTM, but English to fix

PR #244, opened 2025-05-21
[16.0][IMP] product_margin_classification: add price actions on product.template list

huguesdk:
make theoretical price actions available on the `product.template` list instead of only on the `product.product` list.


legalsylvain (approved):
Thanks !

remytms (approved):
LGTM

PR #245, opened 2025-06-13
[16.0][IMP] product_margin_classification: improve taxes computation

huguesdk:
*   add support for taxes of amount_type "fixed".
*   take the `include_base_amount` field into account.

### DIFF
--- a/product_margin_classification/models/product_product.py
+++ b/product_margin_classification/models/product_product.py
@@ -76,8 +76,9 @@ class ProductProduct(models.Model):
     ):
         precision = self.env["decimal.precision"].precision_get("Product Price")
         if classification:
-            multi = (100 + classification.markup) / 100
-            if sale_taxes.filtered(lambda x: x.amount_type != "percent"):
+            base_price = standard_price * ((100 + classification.markup) / 100)
+            extra_included_taxes = 0
+            if sale_taxes.filtered(lambda x: x.amount_type not in ("percent", "fixed")):
                 raise ValidationError(
                     _(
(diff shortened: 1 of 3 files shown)


product_margin_classification/models/product_product.py
```diff
@@ -87,10 +88,17 @@ def _get_margin_info(
                     % (product_name)
                 )
             for tax in sale_taxes.filtered(lambda x: x.price_include):
-                multi *= (100 + tax.amount) / 100.0
+                # this code assumes that taxes having price_include set to
+                # True and of amount_type "percent" have also
+                # include_base_amount set to True, while those of amount_type
+                # "fixed" have include_base_amount set to False.
```
legalsylvain: Better to write a test and raise a not implemented error, if conditions are not checked.

Also, i'm not sur it will work if there are fixed and percent taxes, in various order.

Dont you think ?

huguesdk: yes, i agree with you. i was hesitating to add a check and raise an error, but since the current code doesn't do it (while already assuming the condition on "percent" taxes), i didn’t do it. i will add it. note that it might break current configurations.

about multiple tax combinations, i added some tests to test the most common cases.

huguesdk: instead of adding exceptions, i improved the code to handle cases with `include_base_amount` `True` or `False`. what do you think?

legalsylvain (approved):
Thanks !

legalsylvain (approved):
/ocabot merge minor

PR #246, opened 2025-07-10
[17.0][ADD] product_margin_report_sale_margin

edescalona:
@BinhexTeam

This new module allows you to add two new fields for the sales margin in the Product Margins report.


christian-ramos-tecnativa (approved):
LGTM! I’d like to add more context to these two new fields. Currently, this report includes all sales and purchases in the range by default when calculating the values.
For example, if I’ve purchased 100 tables but only 10 have been sold so far, the margin will appear negative, which can be misleading. In this case, the report seems to reflect more of a cost vs. revenue view of the product rather than actual profitability.
Here’s a screenshot to illustrate the scenario:
image
This new values add a lot of value to this report.
Thanks

product_margin_report_sale_margin/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module adds 2 new fields for calculating the Sales margin.
```
rrebollo, line 1: You could enrich this with some insides from https://github.com/OCA/margin-analysis/pull/246#pullrequestreview-3006603536

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

adasatorres (approved):
Code review, LGTM!

edescalona:
ping @legalsylvain

PR #247, opened 2025-08-05
[17.0][ADD] product_margin_report_landed_costs

edescalona:
@BinhexTeam

Add the landed costs associated with the product to the Product Margins report.


product_margin_report_landed_costs/tests/test_product_margin_report_landed_costs.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2025 Binhex <https://www.binhex.cloud>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from unittest.mock import patch
+
+from odoo.tests.common import TransactionCase
```
rrebollo:
```suggestion
from odoo.addons.base.tests.common import BaseCommon
```
`BaseCommon` is the OCA-recommended approach for tests from 17.0 onward.

product_margin_report_landed_costs/tests/test_product_margin_report_landed_costs.py
```diff
@@ -0,0 +1,73 @@
...
+
+from odoo.addons.product_margin.models.product_product import ProductProduct
+
+
+class TestProductMarginAccumulatedLandedCost(TransactionCase):
```
rrebollo:
```suggestion
class TestProductMarginAccumulatedLandedCost(BaseCommon):
```

rrebollo (approved):
Pure Code Review. LGTM!!! I do think functional review is critical on this kind of addons. Great contribution.

product_margin_report_landed_costs/tests/test_product_margin_report_landed_costs.py
```diff
@@ -0,0 +1,73 @@
...
+                "date_to": "2022-12-31",
+            }
+        }
+        fake_accumulated_cost = 100.0
+        StockValuationAdjustmentLines = self.env["stock.valuation.adjustment.lines"]
+        with patch.object(
+            ProductProduct,
+            "_compute_product_margin_fields_values",
+            return_value=super_result,
+        ):
```
rrebollo, line 47: Really clever technique! I'm curious to learn more: What does it do, how does it work, why is it needed, and when would I use it?

rrebollo:
I don't have extensive knowledge of stock valuation methodologies, but I am aware of the key differences between Anglo-Saxon and Continental accounting principles. Perhaps you could mention any prerequisites regarding those accounting standards in addon's documentation.

product_margin_report_landed_costs/tests/test_product_margin_report_landed_costs.py
edescalona: Hi @rrebollo , I'll answer your questions, and you can let me know if you understand my answers.

What does it do?
It simulates fixed values ​​in the methods (`_compute_product_margin_fields_values`, `_read_group`). One detail in the case of `patch.object `to the `_compute_product_margin_fields_values` ​​method: since it's an inheritance, the class that must be referenced is the one with the original method, that is, in the path (`from odoo.addons.product_margin.models.product_product`) of the `ProductProduct` class. If you see that the path is with the `product_margin` module, it does.

How does it work?
What it does is that instead of executing the logic of the method (`_compute_product_margin_fields_values`) indicated in `patch.object`, it will always return the value of super_result, thus ensuring that the values ​​are fixed.

Why is it necessary?
In this case, what I needed was to work with the values ​​returned by the method to perform some calculations, and well, that was the most direct option. There are always other ways to perform the same test, but it worked for me in this case.

When to use it?
Basically, when you need to set values ​​and don't want the entire method logic to run for that specific test, since it might have been verified in another test.

edescalona:
ping @rrebollo

rrebollo:
Nice job @edescalona

edescalona:
ping @Christian-RB

PR #251, opened 2025-11-11
[16.0][IMP]product_margin_classification: add search function

mihien:


legalsylvain:
Hi. Thanks for your contribution. Could you add a test ?

legalsylvain (approved):
Makes sense.

Thanks for this patch.

/ocabot merge patch

PR #252, opened 2025-11-24
[18.0][MIG] sale_margin_delivered_security: Migration to 18

Andrii9090-tecnativa:
Straightforward migration to v18

https://www.loom.com/share/cd61599fbe834ce8ae79431960dcb935

@moduon MT-11820


sale_margin_delivered_security/models/sale_order_line.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2024 Moduon Team S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+from odoo import fields, models
+
+
+class SaleOrderLine(models.Model):
+    _name = "sale.order.line"
+    _inherit = ["sale.order.line", "product.cost.security.mixin"]
```
chienandalu: Yo don't really need to mix it in again 
```suggestion
    _inherit = "sale.order.line"
```

chienandalu (approved):
Thanks :)

Gelojr (approved):
Congratulations for the work on this module @Andrii9090 , the PR is approved.
The following functional tests were executed:

- Test 1: A user with cost-view permissions successfully viewed the delivered-margin fields on a sales order line after delivery; all values were correctly computed and displayed.
- Test 2: A user without cost permissions could not see any cost or margin fields, including both the delivered-margin fields added by the module and the standard Odoo cost and margin fields; all of them were correctly hidden.
- Test 3: A user without cost permissions could not access the delivered-margin fields even in developer mode; the fields were not available in the view definition.
- Test 4: Delivered-margin values were correctly computed after a partial delivery and were only visible to users with cost-view permissions; users without permissions did not see any margin-related fields.
- Test 5: After removing cost-view permissions, the user no longer saw any delivered-margin fields on existing sales order lines, confirming that visibility reacts immediately to permission changes.
- Test 6: A user with cost-view permissions could see the product cost on the product form and the delivered-margin fields on sales order lines, all correctly shown as read-only.
- Test 7: Editing the standard cost on the sales order line after delivery may cause inconsistencies.
A user with cost-edit permissions can correctly modify the product cost on the product form, and new sales orders will use that updated value. However, this user can also manually edit the standard Odoo cost field on the sales order line itself. If a delivery has already been completed, changing this cost afterwards creates an inconsistency because the delivered-margin values were calculated using the previous cost. As an improvement, it would be helpful for Odoo to automatically log a message in the chatter whenever the standard cost on the sales order line is manually modified, so that any post-delivery cost changes remain traceable.

rafaelbn (approved):
Functional review 👍🏼 

/ocabot merge patch

PR #263, opened 2026-05-04
[18.0][ADD] loyalty_margin_computation: New module

Andrii9090-tecnativa:
It is a base module that only adds a field to `loyalty.reward` to define a formula for margin calculation

@pedrobaeza can you review please?

TT62183
@Tecnativa

PR #262, opened 2026-04-28
[18.0][ADD] sale_margin_pricelist_computation: New module

Andrii9090-tecnativa:
This module extends product pricelists to support Python-based formulas 
for computing product cost and product sale price for margin calculations.
image

@pedrobaeza can you review it?

@Tecnativa
TT62183

PR #264, opened 2026-05-04
[18.0][ADD] sale_loyalty_margin_computation: New module

Andrii9090-tecnativa:
Depends:

- [ ] #262 
- [ ] #263 

@pedrobaeza ping

TT62183
@Tecnativa


sale_loyalty_margin_computation/models/loyalty_reward.py
```diff
@@ -0,0 +1,61 @@
...
+from odoo.exceptions import ValidationError
+from odoo.tools.float_utils import float_compare
+from odoo.tools.safe_eval import safe_eval
+
+
+class LoyaltyReward(models.Model):
+    _inherit = "loyalty.reward"
+
+    def _get_sale_margin_formula_eval_context(self):
+        main_product = self.env["product.product"].new(
```
pedrobaeza, line 15: Why are you using here a fake register? That should be done only for the check, but don't create a general method that do this. Put it as a inner method, or inside the check itself.

pedrobaeza (changes requested):

pedrobaeza (approved):

sale_loyalty_margin_computation/models/sale_order_line.py
```diff
@@ -0,0 +1,39 @@
...
+                str(line.reward_id.sale_margin_formula).strip(),
+                eval_context,
+                mode="exec",
+                nocopy=True,
+            )
+            origin_lines[0].margin = eval_context.get("result", 0)
+            origin_lines[0].margin_percent = (
+                origin_lines[0].margin / origin_lines[0].price_subtotal
+            )
+            break
```
cristina-hidalgo-tecnativa: I'm not sure if this case can happen in practice, but what would occur if two loyalty programs were active at the same time, each with its own `sale_margin_formula`? The `break` would exit the loop after the first reward, so the second program's formula would never run.

I understand the `break` prevents applying the same formula multiple times when a multi_gift reward generates several lines (all sharing the same `reward_id`). But wouldn't skipping by already-processed `reward_id` be safer? 

Something like this:

    processed_reward_ids = set()
    for line in self.filtered(...):
        if line.reward_id.id in processed_reward_ids:
            continue
        processed_reward_ids.add(line.reward_id.id)
        ...

cristina-hidalgo-tecnativa (commented):
Locally tested since runboat appears red. 

I'm not sure if this is intentional, but I created a multi-gift promotion with 2 gifts and a formula `result = origin_lines[0].margin + 10`. 

Image

After applying the reward, the margin on the origin line was `22.40`

Image

Shouldn't it be 12.40? Since `sale_margin_formula` is defined at the reward level (not per gift line), I'd expect the formula to apply once per reward, regardless of how many gift lines it generates.

Andrii9090-tecnativa:
@cristina-hidalgo-tecnativa I tested this, and it seems to calculate correctly.
Unit price of the original product: 15.8
Total: 31.6
Cost: 14.00
Margin: 3.60
After applying the formula, the margin is set to 13.60.
Could you please provide more details about your test cases?

image

cristina-hidalgo-tecnativa:
> @cristina-hidalgo-tecnativa I tested this, and it seems to calculate correctly. Unit price of the original product: 15.8 Total: 31.6 Cost: 14.00 Margin: 3.60 After applying the formula, the margin is set to 13.60. Could you please provide more details about your test cases?
> image

It seems our result of 22.40 was indeed transient — on a subsequent order with the same configuration we got the expected 12.40.

That said, I also tested with two multi-gift programs applied to the same quotation, each with their own `sale_margin_formula` (+10 and +5).

imagen

imagen


And I obtained the following margin:

imagen


Shouldn't it be 17.40 (2.40 + 10 + 5) ?

pedrobaeza (commented):
Not sure if for consistency with the rewards not having a formula, we should as well transfer the margin of the reward line to the main one.

Andrii9090-tecnativa:
@cristina-hidalgo-tecnativa Have you tested it with the changes I pushed yesterday?

Andrii9090-tecnativa:
@pedrobaeza After yesterday's changes, when a reward has a formula configured, the margin of each reward line is now set to 0. Then, the result of the margin formula is applied to one of the origin lines.
And now each formula is applied only once per reward, regardless of whether it is a single gift or a multi-gift reward.

pedrobaeza:
Yes, thanks, Andrii, please check what do you think of https://github.com/OCA/margin-analysis/pull/264#pullrequestreview-4233797700

cristina-hidalgo-tecnativa:
> @cristina-hidalgo-tecnativa Have you tested it with the changes I pushed yesterday?

Yes, tested this morning. Single multi-gift formula looks fixed. But the two-programs case still fails — with +10 and +5 formulas we get 22.40 instead of 17.40.

victoralmau:
Could you rebase and remove test-requirements.txt commit?

PR #267, opened 2026-05-18
[16.0][IMP] account_invoice_margin: set new option 'optional' hide to margin fields on tree views.

legalsylvain:
**Rational**: 
- the margin fields are usefull to be in the tree. However, it is not usefull for all user in a day to day basis. put optional='hide' to make the invoice and invoice lines list tree view lighter.

- prevent to display useless fields on account.move.lines (when not related to out invoices)

- do not display margin in move tree views, if there are not out invoices


legalsylvain:
@@sergio-teruel could you take a look on this one ? 

thanks !

sergio-teruel (approved):
That sounds perfect to me

legalsylvain:
@sergio-teruel  could you merge this one ? 

thanks !

legalsylvain:
Weird error ... nothing to do with the current PR.

@sergio-teruel : I tried to fix the problem in https://github.com/OCA/margin-analysis/pull/267/changes/1aa08374f9afc857e3465ac87bac2550fe591918

could you rerun the merge process ?

thanks !

Raul-S73:
Hi, thanks for the contribution. I wanted to share an effect we are seeing after this change: several customers have reported that the Margin field has disappeared from vendor bill lines.

We understand the intention of not showing fields that may seem irrelevant on certain documents, but hiding the column with `column_invisible` only affects the UI: the margin is still computed internally. In practice, the computation cost remains while users who need the information lose visibility of it.

https://github.com/odoo/odoo/blob/6f6477a5fa84bcab9c1e740b3d5a32683f3b7191/addons/account/models/account_move.py#L4053

https://github.com/OCA/margin-analysis/blob/3035394a3b8bcf3c4f2c5e34a5bf0ea61ceefeb5/account_invoice_margin/models/account_invoice.py#L40

For that reason, we believe this change does not address the underlying issue and introduces a significant usability side effect. Would it be possible to keep the field available on vendor bills as well, perhaps by relying on optional columns or user permissions instead of hiding it globally?

In the meantime, we have had to override this behavior in a custom module.

Issue #268, opened 2026-06-01
[16.0] product_standard_margin: Push to pypi


houzefa-abba:
Hi the 16.0 version of [product_standard_margin](https://github.com/OCA/margin-analysis/blob/16.0/product_standard_margin) is not the latest [on pypi](https://pypi.org/project/odoo-addon-product-standard-margin/#history), could we refresh please?

I'd like to get translations [added in 2025](https://github.com/OCA/margin-analysis/commits/16.0/product_standard_margin).

Thanks
@legalsylvain @sbidoul

sbidoul:
I found this in the bot log for `account_invoice_margin`, which likely interrupts the nightly publisher job for this repo before it can publish `product_standard_margin`:

```
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   | [2026-06-01 03:35:48,524: ERROR/ForkPoolWorker-1] command ['twine', 'check', 'odoo_addon_account_invoice_margin-16.0.1.0.1.3-py3-none-any.whl'] in /tmp/tmpspxerm_n failed with ret>
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   | Checking odoo_addon_account_invoice_margin-16.0.1.0.1.3-py3-none-any.whl: FAILED
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   | ERROR    `long_description` has syntax errors in markup and would not be
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   |          rendered on PyPI.
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   |          line 37: Warning: Cannot scale image!
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   |            Could not get size from
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   |          "https://raw.githubusercontent.com/OCA/margin-analysis/16.0/account_inv
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   |          oice_margin/static/description/account_invoice_form.png":
Jun 01 05:35:48 ax52.odoo-community.org oca-github-bot[1886]: worker-1   |            Requires Python Imaging Library.
[...]

PR #269, opened 2026-06-01
[FIX] account_invoice_margin: error when deploying on pypi.

legalsylvain:
See: https://github.com/OCA/margin-analysis/issues/268


legalsylvain:
hi @houzefa-abba could you take a look on this one, and approve it, if all is OK ?

houzefa-abba (approved):
Seems all good to me, we lose alt-texts on 2 images but worth it if it fixes pypi uploads. :)

PR #271, opened 2026-06-15
[19.0] Update copier template to v1.43

pedrobaeza:
@Tecnativa


pedrobaeza:
The error is because there's no modules yet (it shouldn't, but it should be fixed at other level).

PR #273, opened 2026-06-18
[16.0][FIX]  copier update from 1.29 to 1.43

legalsylvain:
copier update from 1.29 to 1.43

Try to fix pre-commit issue, bumping copier update : 
Ex : https://github.com/OCA/margin-analysis/pull/272


legalsylvain:
hi @OCA/accounting-maintainers Could you take a look and merge this PR that fixes current red CI ? 

thanks !

sergiocorato (approved):

pedrobaeza:
I think you should separate the fix from the copier update. If not, launching ocabot with patch will raise a lot of module versions without need, and putting nobump will have no raise in the affected module. And are you sure that README change will not be undone by the bot? It has its own special version for doing that, so the other possibility is just to not change READMEs at all.

sale_margin_delivered/reports/sale_report.py
```diff
@@ -11,10 +11,10 @@ class SaleReport(models.Model):
 
     def _select_additional_fields(self):
         res = super()._select_additional_fields()
-        res[
-            "margin_delivered"
-        ] = f"""SUM(l.margin_delivered
-            / {self._case_value_or_one('s.currency_rate')}
-            * {self._case_value_or_one('currency_table.rate')})
-        """
+        res["margin_delivered"] = (
```
sergiocorato: This part in v. 18.0 is the same and it doesn't throw errors, could you split this change in another PR and commit only copier update?

sergiocorato (changes requested):

sale_margin_delivered/reports/sale_report.py
legalsylvain: ~no. This line generate a E221 error.~

Done

legalsylvain:
> I think you should separate the fix from the copier update. If not, launching ocabot with patch will raise a lot of module versions without need, and putting nobump will have no raise in the affected module. And are you sure that README change will not be undone by the bot? It has its own special version for doing that, so the other possibility is just to not change READMEs at all.

indeed ! done.

pedrobaeza (approved):
Let's merge this directly, as it doesn't need ocabot.

PR #278, opened 2026-08-10
[18.0][FIX] sale_margin_pricelist_computation: Use sale.order.line _get_margin_pricelist_eval_context to test pricelist.item eval context

Andrii9090-tecnativa:
This change allows `_get_margin_pricelist_eval_context` to be overridden from other modules, making it possible to extend the evaluation context for each margin formula and evaluate the formula directly from `pricelist.item`.

@pedrobaeza can you review please?

@Tecnativa

### DIFF
--- a/sale_margin_pricelist_computation/models/product_pricelist_item.py
+++ b/sale_margin_pricelist_computation/models/product_pricelist_item.py
@@ -1,6 +1,5 @@
 from odoo import Command, api, fields, models
 from odoo.exceptions import ValidationError
-from odoo.tools.float_utils import float_compare
 from odoo.tools.safe_eval import safe_eval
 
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
This is going to fail if the formula includes a division by one of the variables.

sale_margin_pricelist_computation/models/product_pricelist_item.py
```diff
@@ -22,11 +21,25 @@ def _check_margin_formula(self):
...
             if rec.margin_cost_price_formula:
-                self._eval_python_code_margin_formula(rec.margin_cost_price_formula)
+                rec._eval_python_code_margin_formula(rec.margin_cost_price_formula)
 
     def _eval_python_code_margin_formula(self, test_string):
+        pricelist = self.env["product.pricelist"].new(
+            {
+                "name": "Test Pricelist",
+            }
+        )
```
pedrobaeza: Remember to remove the trailing comma in these cases for saving lines.

```suggestion
        pricelist = self.env["product.pricelist"].new(
            {"name": "Test Pricelist"}
        )
```

Andrii9090-tecnativa:
> This is going to fail if the formula includes a division by one of the variables.

Using dummy data to simulate a real order should avoid this error.

pedrobaeza (approved):
We can fast-track it.

/ocabot merge patch

